PR #19, opened 2025-02-26
Test 6.1.35: Add validation for contradicting remediations in CSAF documents

milux:
Introduced test "6.1.35" to validate and detect contradicting remediation categories for the same product. Added traits and implementations to abstract remediation and vulnerability handling for easier extensibility across CSAF versions. Updated test infrastructure and presets to include the new validation logic.


milux:
The Spec states that "This takes indirect relations through product groups into account.".
The current state of this test doesn't perform any product group resolution.
I'm not sure how this is to be understood?

milux:
My implementation currently considers all pairs of different categories as error, but this is invalid. Converting back to draft for now.

csaf-lib/src/csaf/csaf2_0/getter_implementations.rs
```diff
@@ -0,0 +1,78 @@
+use std::collections::BTreeSet;
+use std::ops::Deref;
+use std::str::FromStr;
+use crate::csaf::csaf2_0::schema::{CommonSecurityAdvisoryFramework, ProductGroup, ProductTree, Remediation, Vulnerability};
+use crate::csaf::csaf2_1::schema::CategoryOfTheRemediation as Remediation21;
+use crate::csaf::getter_traits::{CsafTrait, ProductGroupTrait, ProductTreeTrait, RemediationTrait, VulnerabilityTrait};
+use crate::csaf::helpers::resolve_product_groups;
+
+impl RemediationTrait for Remediation {
```
oxisto, line 9: Should we move this file to a more "neutral" place? It is a little bit confusing that we now import csaf 2.1 schema things into a csaf 2.0 folder.

Also needs a documentation

csaf-lib/src/csaf/csaf2_0/getter_implementations.rs
```diff
@@ -0,0 +1,78 @@
...
+                if let Some(product_ids) = resolve_product_groups(doc, product_groups) {
+                    product_set.extend(product_ids.iter().map(|p| p.to_string()));
+                }
+            }
+            Some(product_set)
+        }
+    }
+}
+
+impl VulnerabilityTrait for Vulnerability {
```
oxisto, line 41: doc pls, goes for the rest of the traits as well.

Cargo.lock
```diff
@@ -319,9 +319,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.89"
+version = "1.0.93"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f139b0662de085916d1fb67d2b4169d1addddda1919e696f3252b740b629986e"
+checksum = "60946a68e5f9d28b0dc1c21bb8a97ee7d018a8b322fa57838ba31cc878e22d99"
```
oxisto, line 324: Can you maybe extract the cargo update + schema update to a separate PR? its hard to review this :(

oxisto (commented):
In general this needs more documentation, especially how the trait magic works and please do the update / schema regen in a separate PR so we can focus on the code here which is non-trivial.

tschmidtb51 (commented):
Please also state the test number in the PR, if it is trying to implement a test.

tschmidtb51:
Just converting back to draft to male it easier for me to review things.

milux:
> ... and please do the update / schema regen in a separate PR...

Then we are lacking test cases and especially **remediation categories**.
We would have to do the schema update first than. Fun. -.-

oxisto:
> > ... and please do the update / schema regen in a separate PR...
> 
> Then we are lacking test cases and especially **remediation categories**. We would have to do the schema update first than. Fun. -.-

ok then let's keep it in this file and hopefully the next tests will be easier.

milux:
> ok then let's keep it in this file and hopefully the next tests will be easier.

Most certainly, because I had to do all the hard lifting here... :joy:

csaf-lib/src/csaf/csaf2_0/getter_implementations.rs
oxisto: > This is a normalization step. Since 2.1 extends the set but keeps identifiers of 2.0, we can do it like this. Otherwise, we would need to directly use `String`s instead of `enum`s or implement our own `enum`, which is both non-ideal...

But that would mean that we accidentally would parse something as "CSAF 2.1" in CSAF 2.0, e.g., if 2.1 has an identifier which 2.0 does not have. Probably it would be better to use 2.0 structs in 2.1 and "extend" them somehow, but not the other way around

csaf-lib/src/csaf/csaf2_1/validation.rs
```diff
@@ -1,28 +1,32 @@
 use super::product_helper::*;
 use super::schema::CommonSecurityAdvisoryFramework;
 use crate::csaf::helpers::find_duplicates;
-use crate::csaf::validation::{Test, Validatable, ValidationPreset};
+use crate::csaf::validation::{test_6_01_35_contradicting_remediations, Test, Validatable, ValidationPreset};
```
tschmidtb51, line 4: Why do we need 6.1.35 explicit, all other not?

csaf-lib/src/csaf/csaf2_1/validation.rs
```diff
@@ -111,4 +116,27 @@ mod tests {
...
+            ("03", "Product CSAFPID-9080702 has contradicting remediations: workaround, fix_planned and optional_patch"),
+            ("04", "Product CSAFPID-9080701 has contradicting remediations: mitigation, fix_planned and optional_patch"),
+        ].iter() {
+            let doc = load_document(format!("../csaf/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-35-{}.json", x).as_str()).unwrap();
+            assert_eq!(
+                Err(format!("{}", err)),
+                test_6_01_35_contradicting_remediations(&doc)
+            )
+        }
+    }
```
tschmidtb51, line 141: Can we have one output line per tested file? This makes it easier to spot if a file is missed.

tschmidtb51 (commented):
Some minor comments, other than that LGTM

csaf-lib/src/csaf/csaf2_1/validation.rs
milux: Until this PR, tests have been copy-pastet for 2.0 and 2.1. This is the first test residing in a common location, using our own Traits. Therefore, it has to be imported here.

csaf-lib/src/csaf/csaf2_1/validation.rs
milux: Sure. Will implement an output of the file under test.

milux: Testing output is already in place, it turns out.
`cargo test -- --nocapture` will output information about parsed docs.
If the scattered output (due to multiple test threads) is an issue: `cargo test -- --nocapture --test-threads 1`

csaf-lib/src/csaf/csaf2_1/validation.rs
milux: @tschmidtb51 It just came to my mind that this poses another interesting question.
Per spec, all test from 6.1.34 onward are CSAF 2.1 only. However, they can still make sense for CSAF 2.0 documents. Max recursion (6.1.34) and contradicting remediations (6.1.35) are good examples for this.
Should it be possible to run these tests for 2.0 documents, at least manually?
Or shall I move the test code to the 2.1 package in case we do not want to allow this?
@oxisto Any further thoughts on this?

oxisto: Technically, since they are only CSAF 2.1 tests, they should only be run on CSAF 2.1 documents, since otherwise we would declare a CSAF 2.0 document invalid because it fails a CSAF 2.1 test.

milux: @oxisto Regarding presets, absolutely agreed! How about manual runs on specific documents?

oxisto (approved):

PR #27, opened 2025-03-21
Refactor validations and improve error reporting

milux:
Separated validations into individual modules for better modularity. Enhanced ValidationError structure for detailed error reporting, including message and instance path. Updated test cases and references to use the refactored structure.


oxisto (approved):
Looks reasonable. In general it would be good if we would start to add a little more documentation to the structs and interfaces.

PR #53, opened 2025-07-11
Tests 6.1.{12,13}, Resource Embedding

milux:
I would suggest to merge this soon. It contains the code to embed external resources which used to be loaded from the repo fs via hardcoded relative paths.


oxisto (commented):
Can we avoid to have this 40k file with language tags in this repo? Where it is coming from? We at least need some attribution and license for this file

PR #68, opened 2025-11-18
feat: validation API

oxisto:
This PR adds a new validation API that should provide a better experience for the `csaf-validator` binary and other users of this library. Core features are:
- Three level of validation using `validate_preset`, `validate_tests`, `validate_by_test`
- All validation functions return a `ValidationResult` containing individual test results and errors

### DIFF
--- a/csaf-rs/src/csaf/validation.rs
+++ b/csaf-rs/src/csaf/validation.rs
@@ -2,6 +2,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::str::FromStr;
+use TestResultStatus::{*};
 
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
 pub struct ValidationError {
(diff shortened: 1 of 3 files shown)


csaf-rs/src/csaf/validation.rs
```diff
@@ -19,7 +20,36 @@ impl std::fmt::Display for ValidationError {
...
+/// Result of a CSAF validation
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ValidationResult {
+    /// Whether the validation was successful (no errors)
+    pub success: bool,
+    /// The detected CSAF version
+    pub version: String,
+    /// List of validation errors (empty if successful)
+    pub errors: Vec<ValidationError>,
```
tziemek: I don't see any reading usage of this field, and it is just cloned from all the errors within `test_results`, so I think we can remove it.

oxisto: I replaced it with a `num_errors` field, which I still think is convinient to have.

PR #155, opened 2025-11-20
Refactor validation functions to collect and aggregate errors

### LINKED ISSUE #148: Extend tests to include multiple error messages
After https://github.com/csaf-rs/csaf/pull/73 is merged, we need to have a close look at the existing tests so that they can return multiple error messages and do not abort at the first error.

milux:
- Reverted some returns back to non-Vec returns in Traits.
- Fixed semantics of `visit_all_products[_generic]` and `visit_branches_rec` functions.
- Fixed semantics of related tests to collect errors into a lazy-init `Option<Vec<...>>` where suitable.

Partially fixes #148

### DIFF
--- a/csaf-rs/src/csaf/csaf_traits.rs
+++ b/csaf-rs/src/csaf/csaf_traits.rs
@@ -45,7 +45,7 @@ pub trait DocumentTrait {
     fn get_tracking(&self) -> &Self::TrackingType;
 
     /// Returns the distribution information for this document with CSAF 2.1 semantics
-    fn get_distribution_21(&self) -> Result<&Self::DistributionType, Vec<ValidationError>>;
+    fn get_distribution_21(&self) -> Result<&Self::DistributionType, ValidationError>;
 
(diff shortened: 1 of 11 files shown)


csaf-rs/src/csaf/csaf_traits.rs
```diff
@@ -75,7 +75,7 @@ pub trait DistributionTrait {
     fn get_tlp_20(&self) -> Option<&Self::TlpType>;
```
tziemek: Just saw this now, why do the `_20` functions not return a `Result` but an `Option`? Can be handled in another PR, just wondering if there is a particular reasoning.

tziemek (approved):
LGTM

csaf-rs/src/csaf/csaf_traits.rs
tschmidtb51: I don't have an opinion on that (yet). I just wanted to point of that there are value changes from TLP v1 to TLP v2...

So happy to leave it as is for now.

PR #163, opened 2025-11-21
feat: add minimum required rust version

### LINKED ISSUE #20: Document minimum Rust version
We should document in the README what Rust version we expect to see, if someone wants to compile it from source.

tziemek:
resolves #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,17 @@
 # csaf-rust
 
-This repository is a proof-of-concept for a CSAF library in Rust that relies on automatically generating CSAF document structs from the JSON schema.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -13,9 +22,6 @@ If you want to build `csaf-validator` on your own, please install Rust (see http
 git submodule init
 git submodule update --remote
 
-# make sure that local assets are in sync with git submodules
```
milux: Isn't this still relevant?

milux: Well, in that case a remote update of the submodules (lines above) wouldn't make sense either, since the two go together.
Question is whether you want to use the very latest version of submodules and their corresponding assets or not. I would say we want developers to always use the latest versions, and then this should stay in there.

milux (approved):

PR #165, opened 2025-11-21
fix: return all validation errors for test 6.1.2

### LINKED ISSUE #75: Implement 6.1.2 Multiple Definition of Product ID

tziemek:
resolve #75


milux (approved):
This should be functionally identical to the previous code since there is no looping and the product visiting already has implicit exhaustive error-collecting semantics.
However, it makes sense as a preparation for the simplification you suggested, so lgtm.

PR #175, opened 2025-11-25
fix: return all validation errors for test 6.1.7

### LINKED ISSUE #80: Implement 6.1.7 Multiple Scores with same Version per Product

tziemek:
resolves #80 

* only breaks early if an invalid CVSSv3 version is found
* splits off gathering of mapping product->metric

### DIFF
--- a/csaf-rs/src/csaf_traits.rs
+++ b/csaf-rs/src/csaf_traits.rs
@@ -445,6 +445,40 @@ pub trait ContentTrait {
     fn get_content_json_path(&self, vulnerability_idx: usize, metric_idx: usize) -> String;
 }
 
+/// Types of vulnerability metrics known until CSAF 2.1
+#[derive(Hash, Eq, PartialEq, Clone)]
+pub enum VulnerabilityMetric {
+    SsvcV1,
+    CvssV2,
+    CvssV3(String),
+    CvssV4,
+    Epss,
+}
+
+/// Display implementation for VulnerabilityMetrics.
+impl Display for VulnerabilityMetric {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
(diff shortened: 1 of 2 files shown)


tschmidtb51:
> only breaks early if an invalid CVSSv3 version is found

Why?

tschmidtb51:
> There is an error produced out of it, but we can't really handle invalid CVSSv3 versions, because we don't have a schema.

For now, my approach would be (but that might change based on https://github.com/oasis-tcs/csaf/issues/1232 - so feel free to implement or skip until a final decision is made):
Is it really necessary to parse CVSS? If the version is not present - yes. But otherwise: `version` is a string. If the product is mentioned in a score with the same version, that should also apply regardless whether we can validate that CVSS v3 version (that's a different test).
Also, we need to make sure the test does not break if an unknown version is inserted along the chain (e.g. `3.2`, `3.0`, `3.0` for the same product should still report the `3.0` double independent of the sorting (`3.2` first vs. last) - I didn't check the code whether it is already implemented that way.)

tziemek:
> For now, my approach would be (but that might change based on [oasis-tcs/csaf#1232](https://github.com/oasis-tcs/csaf/issues/1232) - so feel free to implement or skip until a final decision is made): Is it really necessary to parse CVSS? If the version is not present - yes. But otherwise: `version` is a string. If the product is mentioned in a score with the same version, that should also apply regardless whether we can validate that CVSS v3 version (that's a different test). Also, we need to make sure the test does not break if an unknown version is inserted along the chain (e.g. `3.2`, `3.0`, `3.0` for the same product should still report the `3.0` double independent of the sorting (`3.2` first vs. last) - I didn't check the code whether it is already implemented that way.)

Ah ok, that clarifies a few cases. I'll change it not to test if the version is known (as this is test 6.1.8). The duplicate cases will then still be covered, regardless if they are within one metric or not (within one vulnerability).

PR #176, opened 2025-11-25
Add test conformance section to README

tziemek:
Added a section for test conformance to the standard with mandatory tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,17 +68,19 @@ csaf-validator --csaf-version 2.1 --test-id 6.1.34 my-csaf-2-1-document.json
 ```
 
 
-## Test-Conformance to the Standard
+## Implementation status in regards to the Standard
 
 * :white_check_mark: Implemented
 * :o: Not applicable
 
 ### Mandatory Tests
 
-| Testspecification | 2.0 | 2.1 (experimental) |
+| Test specification | 2.0 | 2.1 (experimental) |
 | --- | --- | --- |
 | 6.1.1 | :white_check_mark:  | |
 | 6.1.2 | :white_check_mark:  | |
 | 6.1.3 | :white_check_mark:  | |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,17 +68,19 @@ csaf-validator --csaf-version 2.1 --test-id 6.1.34 my-csaf-2-1-document.json
...
 
 * :white_check_mark: Implemented
 * :o: Not applicable
 
 ### Mandatory Tests
 
-| Testspecification | 2.0 | 2.1 (experimental) |
+| Test specification | 2.0 | 2.1 (experimental) |
 | --- | --- | --- |
 | 6.1.1 | :white_check_mark:  | |
```
oxisto: Checks are automatically implemented for both 2.0 and 2.1

README.md
```diff
@@ -68,17 +68,19 @@ csaf-validator --csaf-version 2.1 --test-id 6.1.34 my-csaf-2-1-document.json
...
 | 6.1.2 | :white_check_mark:  | |
 | 6.1.3 | :white_check_mark:  | |
 | 6.1.4 | :white_check_mark:  | |
 | 6.1.5 | :white_check_mark:  | |
+| 6.1.6 | | |
+| 6.1.7 | | |
```
oxisto: 6.1.12-13 and 34-49 are implemented but some of the SSVC are correctly disabled. @milux which ones are they?

README.md
tziemek: We can't generalize this, as there are different testcases for both versions.

README.md
tziemek: I wouldn't put them in right now, because they only show the first error. If that's fine for these cases (haven't looked at them) then ok, but we have issues open to look at every test for 2.0 anyway. If that results in only checking if everything is fine and updating this page, that's also ok.

PR #179, opened 2025-11-25
chore: remove `csaf` subfolder

### LINKED ISSUE #178: Move contents of `csaf` create directly inside `csaf-rs` package
For some reason (I honestly don't remember why), we have a `csaf` create (inside `src/csaf`) inside the `csaf-rs` "package". This seems unnecessary and produces types like `csaf_rs::csaf::validation::ValidationResult`. We should instead get rid of the `csaf` intermediate folder.

oxisto:
This PR removes the `csaf` subfolder and also names the *library* `csaf`, so we finally have some nice types such as `csaf::validation::ValidationResult` instead of `csaf_rs::csaf::validatation::ValidationResult`

Fixes #178


oxisto:
> @oxisto Is that allowed? On crates.io, we are still under csaf-rs, correct? (Or did you silently takeover `csaf`?) 🤔

Yes that's allowed. It can be configured, just like the binary name https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-name-field

And yes the package is still csaf-rs, but it does not need to match. In fact it does not even match now since the default name replaces dashes with underscores and our library name is currently `csaf_rs`

tziemek (approved):

PR #214, opened 2025-12-09
feat: add test 6.1.26

### LINKED ISSUE #99: Implement 6.1.26 Prohibited Document Category Name

peinjoh:
Resolves #99

### DIFF
--- a/csaf-rs/src/validations/test_6_1_27_08.rs
+++ b/csaf-rs/src/validations/test_6_1_27_08.rs
@@ -35,7 +35,7 @@ const PROFILE_TEST_CONFIG: DocumentCategoryTestConfig =
 fn test_6_1_27_08_err_generator(document_category: &CsafDocumentCategory, vuln_path_index: &usize) -> ValidationError {
     ValidationError {
         message: format!(
-            "Document with category '{document_category}' must provide at at least either cve or ids  in each vulnerability"
+            "Document with category '{document_category}' must provide at least either cve or ids in each vulnerability"
         ),
         instance_path: format!("/vulnerabilities/{vuln_path_index}/product_status"),
     }


csaf-rs/src/csaf_traits.rs
```diff
@@ -107,38 +108,111 @@ pub trait DocumentReferenceTrait {
...
+    /// Normalizes the document category string by removing leading "csaf" and any whitespace, hyphen or underscore
+    pub fn normalize(&self) -> String {
+        // Regex to match leading "csaf" and any whitespace, hyphen or underscore
+        // before the "csaf" or otherwise anywhere in the string
+        // ^[\s_-]*csaf - matches any leading whitespaces, hyphens or underscores followed by "csaf"
+        // |            - or
+        // [\s_-]      - matches any whitespace, hyphen or underscore anywhere in the string
+        let normalization_regex: Regex = Regex::new(r"^[\s_-]*csaf_|[\s_-]").unwrap();
+        normalization_regex
+            .replace_all(&self.to_string().to_lowercase(), "")
```
tziemek: We should check if this is faster without a regex. Maybe just note it down for later, so we don't forget.

csaf-rs/src/validations/test_6_1_26.rs
```diff
@@ -0,0 +1,112 @@
...
+        return Err(vec![test_6_1_27_6_err_generator_starts_with_csaf(
+            &doc_category,
+            version,
+        )]);
+    }
+
+    // throw error if document category is too similar to known categories
+    // this is done by comparing normalized versions of the categories
+    for normalized_known_category in DocumentCategory::known_profiles_normalized(version) {
+        println!(
```
tziemek: This should rather be a debug log.

csaf-rs/src/csaf_traits.rs
peinjoh: It most likely is, will do.

PR #224, opened 2025-12-12
feat: add test 6.1.27.10

### LINKED ISSUE #109: Implement 6.1.27.10 Action Statement

peinjoh:
resolves #109


csaf-rs/src/validations/test_6_1_27_10.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    let mut errors: Option<Vec<ValidationError>> = None;
+    // for each vulnerability
+    for (v_i, vulnerability) in vulnerabilities.iter().enumerate() {
+        // generate hashmap of all known_affected product or group ids with value of known_not_affected path index
+        let mut known_affected_product_or_group_ids: HashMap<String, usize> = HashMap::new();
+        if let Some(product_status) = vulnerability.get_product_status() {
+            if let Some(known_affected) = product_status.get_known_affected() {
+                for (kna_i, known_affected_entry) in known_affected.into_iter().enumerate() {
+                    if known_affected_product_or_group_ids.contains_key(known_affected_entry) {
```
tziemek: Same here, I think we shouldn't panic as this should be covered by 6.1.6.

tziemek (changes requested):

csaf-rs/src/validations/test_6_1_27_10.rs
peinjoh: https://github.com/csaf-rs/csaf/pull/223/files/211d4f7e3bf3b253fbba56a40754ec9ab6736844#r2614066201

> You are right, although 6.1.6 covers contradicting status groups (i.e. known_affected vs. known_not_affected).
> 
> Since we're only looking at known_not_affected here, the uniqueness is guaranteed through the schema by `"uniqueItems": true` on the `#/$defs/products_t`. Will remove the check.

Same issue but for `known_affected`, same fix.

tziemek (approved):

tziemek (approved):

PR #231, opened 2026-01-02
feat: profile test category check macro

peinjoh:
This PR: 
* adds a typesafe macro to deduplicate the "is document csaf version X and document category Y" needed for all profile tests
* uses that macro in the profile tests 6.1.27.1 - 6.1.27.11

The categories the tests need to run for can be defined by the ```csaf20``` ```csaf21``` and ```all``` fields of the macro. ```all``` is a shorthand to add the category to both vectors. Due CSAF 2.1 mostly extending CSAF 2.0 here, only ```all``` and ```csaf21``` has been used so far.

**Why is this in draft?**

If we go down the road of adding macros for cases like this, we'll need to publish this macro create to crates.io, too. This is still up for debate, discussion on this is welcome. Because of this, the PR is drafted for now.


peinjoh:
@tziemek and I discussed this further and decided against the usage of proc macros for cases like this. I'll reimplement this without macros.

PR #239, opened 2026-01-15
feat: add test 6.2.1,3,4,7,10,12,15

### LINKED ISSUE #117: Implement 6.2.1 Unused Definition of Product ID

peinjoh:
resolves #117 , #118 , #119 , #120 , #123 , #126, #128 , #131

### DIFF
--- a/csaf-rs/src/validations/test_6_1_27_05.rs
+++ b/csaf-rs/src/validations/test_6_1_27_05.rs
@@ -1,5 +1,5 @@
 use crate::csaf_traits::{CsafTrait, DocumentCategory, DocumentTrait, VulnerabilityTrait};
-use crate::document_category_test_helper::ProfileTestConfig;
+use crate::document_category_test_helper::DocumentCategoryTestConfig;
 use crate::validation::ValidationError;
 
 /// 6.1.27.5 Vulnerability Notes
--- a/csaf-rs/src/validations/test_6_1_27_07.rs
+++ b/csaf-rs/src/validations/test_6_1_27_07.rs
@@ -1,5 +1,5 @@
 use crate::csaf_traits::{CsafTrait, DocumentCategory, DocumentTrait, ProductStatusTrait, VulnerabilityTrait};
-use crate::document_category_test_helper::ProfileTestConfig;
+use crate::document_category_test_helper::DocumentCategoryTestConfig;
 use crate::validation::ValidationError;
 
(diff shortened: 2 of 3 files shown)


csaf-rs/src/validations/test_6_2_01.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// 6.2.1 Unused Definition of Product ID
+///
+/// All defined product IDs need to be referenced at least once in the document.
+pub fn test_6_2_01_unused_definition_of_product_id(doc: &impl CsafTrait) -> Result<(), Vec<ValidationError>> {
+    let mut errors: Option<Vec<ValidationError>> = None;
+
+    // Skips the test for profile "Informational Advisory"
+    if doc.get_document().get_category() == DocumentCategory::CsafInformationalAdvisory {
```
tziemek: Out of consistency, I would use our `PROFILE_TEST_CONFIG` here.

csaf-rs/src/validations/test_6_2_10.rs
```diff
@@ -0,0 +1,47 @@
+use crate::csaf_traits::{CsafTrait, DistributionTrait, DocumentTrait};
+use crate::validation::ValidationError;
+use std::sync::LazyLock;
+
+/// 6.2.10 Missing TLP label
+///
+/// `/document/distribution/tlp/label` must be set.
+pub fn test_6_2_10_missing_tlp_label(doc: &impl CsafTrait) -> Result<(), Vec<ValidationError>> {
+    // We just need to consider get_distribution_20 / get_tlp_20 here, in CSAF 2.1 this field is mandatory and
+    // validated via the schema. This test will not run for CSAF 2.1 documents.
```
tziemek: We should think about what error messages are produced by the schema validation. I guess they won't look like our custom error messages here.

tziemek (changes requested):

csaf-rs/src/validations/test_6_2_10.rs
peinjoh: I don't think they need to. It's not like we're using the schema validation to do this test for CSAF 2.1 docs. This test does not exist in the context of CSAF 2.1, for the reason that a document without a distribution / tlp label is syntactically wrong. I would expect this error to look like any other syntax-related error in CSAF 2.1 docs.

PR #240, opened 2026-01-15
feat: add test 6.2.8,9

### LINKED ISSUE #124: Implement 6.2.8 Use of MD5 as the only Hash Algorithm

peinjoh:
resolves #124 #125 

also: migration from custom from_x functions to From<T> for custum enums HashAlgorithms, DocumentCategory and VersionNumber

### DIFF
--- a/csaf-rs/src/validations/test_6_2_08.rs
+++ b/csaf-rs/src/validations/test_6_2_08.rs
@@ -30,7 +30,7 @@ pub fn test_6_2_08_use_of_md5_as_only_hash_algo(doc: &impl CsafTrait) -> Result<
 
 fn create_md5_only_hash_error(path: &str, hash_index: usize) -> ValidationError {
     ValidationError {
-        message: "hashes product identification helper uses MD5 as the only hash algorithm".to_string(),
+        message: "Product identification helper uses hashes with `md5` as the only hash algorithm".to_string(),
         instance_path: format!("{path}/product_identification_helper/hashes/{hash_index}/file_hashes",),
     }
 }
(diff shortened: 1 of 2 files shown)


csaf-rs/src/csaf_traits.rs
```diff
@@ -1201,15 +1207,134 @@ pub trait HashTrait {
...
+    Sm3,
+    Ssl3Md5,
+    Ssl3Sha1,
+    Whirlpool,
+    Other(String),
+}
+
+impl From<&String> for HashAlgorithm {
+    fn from(algo: &String) -> Self {
+        match algo.as_str() {
```
tziemek: Should we add a `to_lower` here? I would put the algorithms capitalized in my json files, if editing manually.

peinjoh: I think this should be clarified / discussed with @tschmidtb51. The standard defines these values to be derived from the `openssl dgst -list` output, which is lowercase. The default value is lowercase, the tests define their weak algos as lowercase. Secvisogram accepts `SHA1` as a valid value, while `sha1` is flagged.

I would argue that parsing / checking should be case-insensitive, as you could also get around the 6.1.25 mandatory test by having two hashes with `sha1` and `SHA1` and the doc would still be valid. But as far as I have seen, this is correct by the standard.

peinjoh: @aschierl-xitaso this might be relevant to your review.

aschierl-xitaso (approved):
LGTM

PR #254, opened 2026-01-22
chore: satisfy inline format

tziemek:


peinjoh (approved):
LGTM, only format string lines.

PR #262, opened 2026-01-26
refactor: CsafDateTime wrapper

peinjoh:
This PR: 
* adds the CsafDateTime wrapper, which encapsulated the raw datetime string and the parsing result (DateTime / CustomError) from parsing the raw string to RFC3339.
* uses this new wrapper across the application

Some usages still have TODO's those will be resolved with the RevisionHistory refactor, which will introduce casting ParsingErrors into ValidationErrors with little overhead add remove the custom revision history code in the CSAF 2.1 tests.

### DIFF
--- a/csaf-rs/src/csaf/types/csaf_datetime.rs
+++ b/csaf-rs/src/csaf/types/csaf_datetime.rs
@@ -191,7 +191,7 @@ mod tests {
             CsafDateTime::Valid(valid) => {
                 assert_eq!(valid.get_raw_string(), "2024-01-15T10:30:00Z");
             },
-            CsafDateTime::Invalid(err) => panic!("DateTime should have been valid, but returned invalid with: {}", err),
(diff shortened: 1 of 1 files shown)


csaf-rs/src/csaf/types/csaf_datetime.rs
```diff
@@ -0,0 +1,293 @@
...
+use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::str::FromStr;
+
+// ============================================================================
+// CsafDateTime Implementation
+// ============================================================================
+
+#[derive(Debug, Clone)]
+pub enum CsafDateTime {
+    Valid(ValidCsafDateTime),
```
tziemek: Do we need the `ValidCsafDateTime` in the valid case? I thought the raw string was only necessary in the invalid case?

PR #394, opened 2026-02-10
ci: add devguard workflow

tschmidtb51:


tziemek:
> > This won't work, as the full workflow is analyzing docker images and we don't have those. I'll update this PR with the correct pipeline steps.
> 
> Thank you!
> 
> Secret exists so you should be good to go.

Could you invite me to the devguard organisation? I need a few other values to call the workflows correctly.

PR #401, opened 2026-02-11
feat: add test 6.2.20

aschierl-xitaso:


tschmidtb51:
> @tschmidtb51 What is the expected behavior for additional properties appearing in `cvss_v4` metrics? Do we want to get both an error from `6.1.8` (CVSS schema validation using `cvss-v4.0.rev.json` which has set `"unevaluatedProperties": false`), and a warning from `6.2.20` (checking for additional properties)? Currently, the test cases look like that (`oasis_csaf_tc-csaf_2_1-2024-6-2-20-02.json` and `oasis_csaf_tc-csaf_2_1-2024-6-1-08-05.json`).

Currently, yes. If the CVSS group accepts the changes suggested in `cvss-v4.0.rev.json`.
However, the TC might change that behavior with swtiching to Revision 2.

> Or should `6.1.8` only complain about the properties that were erroneously present in 4.0 and have been removed in Revision 1?

No. 6.1.8 complains about anything not compliant with the respective schema.

tschmidtb51:
We hang in the limbo here: We pointed out to the CVSS folks that the change to the new JSON schema is crucial for us - also as the old values floated around and some implementations might still be carrying those. Therefore, we expect the change to come but it is not there yet.
However, if it does not, we will still keep the test files you mentioned and extent the test description about those `cvss_v4` properties...

> And maybe later, test `6.2.20` is changed so that it no longer complains about additional properties in `cvss_v4`, so that `oasis_csaf_tc-csaf_2_1-2024-6-2-20-02.json` is non-faling w.r.t. test `6.2.20`.

This would probably something that would be resolved by filtering those out that are reported through a schema check. Let's see what happens in https://github.com/oasis-tcs/csaf/issues/1288 :face_with_peeking_eye:

tziemek (approved):

peinjoh:
Resolved #401

peinjoh:
Resolves #136

PR #451, opened 2026-03-02
feat: add test 6.1.20 for CSAF 2.1

### LINKED ISSUE #353: Implement 6.1.20 Non-Draft Document Version for CSAF2.1

tziemek:
resolves #353


csaf-rs/src/validations/test_6_1_20.rs
aschierl-xitaso: The path should be `document/tracking/version` (also below)

PR #454, opened 2026-03-02
feat: add test 6.1.22 for CSAF 2.1

### LINKED ISSUE #355: Implement 6.1.22 Multiple Definition in Revision History for CSAF2.1

tziemek:
resolves #355


csaf-rs/src/validations/test_6_1_22.rs
aschierl-xitaso: I think we no longer check for strings here, but for parsed revision numbers.

aschierl-xitaso (approved):

PR #470, opened 2026-03-12
feat: add test 6.1.27.17 for CSAF 2.1

tziemek:


csaf-rs/src/validations/test_6_1_27_17.rs
```diff
@@ -0,0 +1,124 @@
...
+        message: "The note has the correct title. However it uses the wrong category.".to_string(),
+        instance_path: format!("/document/notes[{note_index}]").to_string(),
+    }
+}
+
+/// 6.1.27.17 Reasoning for withdrawal
+///
+/// This test only applies to documents with `/document/category` with value `csaf_withdrawn` and only if the document language is English (i.e., `/document/lang` with value `en`) or unspecified.
+///
+/// Exactly one item in /document/notes[] must have a title `Reasoning for Withdrawal` with category `description` with .
```
peinjoh: Also, this phrasing is a bit different from the prose and the implementation. 

There can only by one note with title "Reasoning for Withdrawal". And that notes category needs to be description. Your text implies that notes with the title and other categories are fine (I would say).

PR #533, opened 2026-04-09
feat: add test 6.2.22

### LINKED ISSUE #308: Implement 6.2.22 Document Tracking ID in Title for CSAF 2.1

peinjoh:
Resolves #308 

This PR:
* adds `/document/title` to the trait
* adds test 6.2.22 for concerning `/document/title` and `/document/tracking/id`
* wires up the test

### DIFF
--- a/csaf-rs/src/validations/test_6_2_22.rs
+++ b/csaf-rs/src/validations/test_6_2_22.rs
@@ -1,0 +1,68 @@
+use crate::csaf_traits::{CsafTrait, DocumentTrait, TrackingTrait};
+use crate::validation::ValidationError;
+
+fn create_tracking_id_in_title_error(title: &str, tracking_id: &str) -> ValidationError {
+    ValidationError {
+        message: format!("The document title '{title}' contains the document tracking id '{tracking_id}'."),
+        instance_path: "/document/title".to_string(),
+    }
+}
+
+/// 6.2.22 Document Tracking ID in Title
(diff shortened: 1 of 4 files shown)


csaf-rs/src/validations/test_6_2_22.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// 6.2.22 Document Tracking ID in Title
+///
+/// It MUST be tested that the /document/title does not contain the /document/tracking/id.
+pub fn test_6_2_22_document_tracking_id_in_title(doc: &impl CsafTrait) -> Result<(), Vec<ValidationError>> {
+    let document = doc.get_document();
+    let title = document.get_title();
+    let tracking_id = document.get_tracking().get_id();
+
+    if title.contains(tracking_id.as_str()) {
```
tziemek, line 19: What about case-insensitivity here? I know the test doesn't state it, but would it make sense?

tziemek (approved):

csaf-rs/src/validations/test_6_2_22.rs
peinjoh: I asked myself the same thing. Probably a question for upstream, but as the tracking IDs are not natural language, and are hopefully only copied and not re-typed, I would think that exact matching should be fine.

Will still open an issue upstream, thanks for the hint!

Issue #409, opened 2026-02-19
Discussion: Return Type Refactor


peinjoh:
As discussed in the last BiWeekly, we need to improve the test return type to accomodate more complex cases.

Currently, all tests return `Result<Ok(()), Err<Vec<ValidationError>>`.

We found the following "cases" of test returns:
- The test `passes`, i.e. was not skipped, all input was valid, the actual check passed
- The test `fails`, i.e. it was not skipped, all input was valid, the actual check failed
- The test is `skipped`, i.e. the test is by definition not applicable, for example:
     - the test only applys to DocumentCategory != Draft, but DocumentCategory == Draft
     - the test only applies to SemVer in revision history and document version , but the document uses IntVer
     - the test is run on the product tree, but the document does not contain a product tree
     - the profile tests (6.1.27.*)
- Some of the data necessary for the test is invalid, i.e. we need to check if the document version contains a pre-release part, but the document version is the invalid value `abcdef`. Here, a `pre-condition failed`, so the test is `non-determinable`.
- The test runs on a larger aggregation, but some of the data in the aggregation is invalid, i.e. a version number in the revision history is `abcdef`, so sorting by revision number is not possible. Here, a `pre-condition failed`, so the test is `non-determinable`. 
- The test runs on multiple "objects" in parralel, of which different objects can `fail` or have their seperate `pre-condition fail`, i.e. the test evaluates 3 cvss objects, one of which is malformed (`non-determinable`) and one of which `fails` the actual test.

Additionally, tests can generate `warnings` (exp: 6.1.15 `source_lang` is present and set, but value is invalid lang tag, `infos` and later, `suggestions` (i.e. quick fixes from the standard).

We discussed this, and would like to propose the following result type:

```
struct ValidationResult {
   result: ResultType  
   warnings: Option<Vec<ValidationWarning>  
   infos: Option<Vec<ValidationInfo>   
}
```

Each test returns a ValidationResult. The result can optionally contain warnings and infos. The "status" of the test is determined by the `result`.

```
type NonDeterminableError = PreconditionError;
enum ResultType {
    Passed
    WasSkipped
    PreconditionFailed(Vec<PreconditionError>>)
    Failed(Option<Vec<NonDeterminable>, Vec<ValidationErrors>)
}
```

To create the `ValidationResult` a builder is provided with the following method:
```
add_info() -> push into infos vec
add_warning() -> push into warnings vec
add_precondition_failed() -> push into precondition vec 
[...]

PR #551, opened 2026-04-10
refactor: add macro for test skip based on doc status

peinjoh:
So far, we have been "hardcoding" the same check in multiple places, always with a comment referencing #409.

This extracts this check into a macro with minimal unit tests. When we do #409, we can just change the return type in one place. I also removed the comments describing this behaviour, as the macro is "speaking" named.

Also, in 6.1.17, I updated the comments to remove the reference to #409. This is not a skip, but a pass of the validation due to an extracted test condition as a guard statement.


csaf-rs/src/validations/test_6_1_38.rs
```diff
@@ -5,7 +5,7 @@ use crate::schema::csaf2_1::schema::LabelOfTlp;
 use crate::validation::ValidationError;
 
 static NON_PUBLIC_SHARING_GROUP_ERROR: LazyLock<ValidationError> = LazyLock::new(|| ValidationError {
-    message: "Document must be public (TLD CLEAR) when using max UUID as sharing group ID.".to_string(),
+    message: "Document must be public (TLP:CLEAR) when using max UUID as sharing group ID.".to_string(),
```
peinjoh, line 8: Unrelated typo fix.

tziemek (approved):

tziemek (approved):

tziemek (approved):

PR #557, opened 2026-04-13
refactor: purl type and tests (6.1.13, 42)

peinjoh:
Resolves some issues around purl / validations around purl / typify.

Typify caused the `pattern` regex on the schema not to be generated, due to it not supporting the `format: uri` property / value. Similarly `format: date-time`, I added an exception to the typegenerator. The regex validation that needed to be done in 6.1.13 is now done during the schema validation / unparsing.

This fixes the previous bug that both CSAF 2.0 and 2.1, having differnt regexes for purl, were using the same one.

As the validation was used by both 6.1.13 and 6.1.42, I added moved the purl validation (without the non-redundant regex validation) into a seperate type with some unit tests.

The the supplemental tests on 6.1.13 that tested the regex working correctly are now unit-tests of `CsafPurl`. I kept them in if we need them as schema tests, with a TODO to fix this one we have schema tests.

I found quite a few more supplemental test cases though, I added. #579 to move these into supplemental test cases.

I also backported the one failing and two passing testcases of 6.1.13 CSAF 2.1 to CSAF 2.0 as suppl test cases.

I also improved the 6.1.42 implementation, which validates that purls in `purls` array differ only in their classifiers. The previous impl did sliding window over the purl items. This worked well for cases like `a -> a -> b -> b`, producing one error message for the change from `a -> b`. For cases like `a -> b -> a -> b -> a`, this would have generated redundant messages for each change. We now calculate the occurance count, and generate error messages for each purl (with classifier removed) that is not the majority.

Also, general improvements to code quality.


csaf-rs/src/csaf/types/purl/csaf_purl.rs
```diff
@@ -0,0 +1,177 @@
...
+use crate::validation::ValidationError;
+use packageurl::PackageUrl;
+use std::ops::Deref;
+use std::str::FromStr;
+
+/// Represents a parsed CSAF PURL that is either valid or invalid.
+#[derive(Debug)]
+pub enum CsafPurl {
+    /// A successfully parsed and validated PURL.
+    Valid(PackageUrl<'static>),
```
tziemek: I think we shouldn't expose the underlying type `PackageUrl` to the public and be implementation agnostic here.

csaf-rs/src/validations/test_6_1_13.rs
```diff
@@ -1,118 +1,28 @@
...
 /// 6.1.13 PURL
-/// Checks the validity of PURLs in the document. Validation is done via a Regex specified in the standard and
-/// via the packageurl::purl parser.
+///
+/// Checks the validity of PURLs in the document. There are different Regex's for the `purl` / `purls` field in CSAF 2.0 and 2.1.
+/// These are enforced during deserialization into the schema types. [CsafPurl] wraps the schema types
+/// and parses the PURL string into a `packageurl::PackageUrl` struct, which performs the actual validation according to the PURL specification.
+///
+/// In this test, we just check if any purls are [CsafPurl::Invalid] and report the errors provided by `packageurl`.
+/// If a purl failed the respective regex, the schema validation failed already, so this test does not run.
```
tziemek: We cannot rely on the schema test having been executed. We currently don't to it for single test calls. Hence we should create an error here even if the schema validation would also throw an error.

tziemek (changes requested):

csaf-rs/src/csaf/types/purl/csaf_purl.rs
peinjoh: Will add indirection layer for PackageUrl Type and wrapper error type. Also double check access to error constructor for tests.

csaf-rs/src/validations/test_6_1_13.rs
peinjoh: This will be a bigger topic during #571 , out of scope here.

csaf-rs/src/csaf/types/purl/csaf_purl.rs
peinjoh: Done, please have another look :)

type-generator/src/main.rs
```diff
@@ -99,7 +99,8 @@ fn fix_2_0_schema(value: &mut Value) {
     for path in fix_paths {
         value.dot_set(path.as_str(), json!({"type": "object"})).unwrap();
     }
-    remove_datetime_formats(value);
+    remove_format(value, "date-time");
+    remove_format(value, "uri");
 }
```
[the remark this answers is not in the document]
peinjoh: We hope this will be replaced soon. Fine for now.

csaf-rs/src/csaf/traits/vulnerabilities/product_ident_helper_trait.rs
```diff
@@ -12,7 +13,22 @@ pub trait ProductIdentificationHelperTrait {
     type HashType: HashTrait;
 
     /// Returns the PURLs identifying the associated product.
```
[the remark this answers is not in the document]
peinjoh: This will be lazy allocated later. Fine for now.

tziemek (approved):

tziemek (approved):

PR #563, opened 2026-04-14
fix: add write permission to publish-crates job so gh-release action can update the release with it's artifacts

tziemek:


oxisto (approved):
Correct, needed according to https://github.com/softprops/action-gh-release?tab=readme-ov-file#permissions

PR #566, opened 2026-04-14
fix: remove writing in build for docs.rs build because they have a readonly file system

tziemek:
When publishing on crates.io a build is triggered to generate the documentation. In our case this build could trigger that files will we updated. Because the used container on crates.io uses a readonly filesystem, this step would break the build and the documentation won't get updated.


peinjoh:
Can you add more context on what the issue was and how this fixes it?

PR #574, opened 2026-04-15
fix: make test 6.1.21 less strict

### LINKED ISSUE #550: Correct 0 or 1 detection in 6.1.21
Interim documents should only have 0 or 1 as first version, but currently checks for 0.

tziemek:
resolves #550


peinjoh (approved):
LGTM

PR #577, opened 2026-04-15
feat: add test 6.3.2

### LINKED ISSUE #138: Implement 6.3.2 Use of CVSS v3.0

### LINKED ISSUE #270: Implement  6.3.2 Use of CVSS v3.0 for CSAF 2.1

peinjoh:
Resolves #138 
Resolves #270 

This PR adds test 6.3.2. 

We have the cvss derserialization now, but I decided against using it, as the overhead ist not justified for what can be achieved with two string compares.

Lets discuss if we should add the following supplemental test cases:
* mix of 3.0 / 3.1 in version / vector  
* version or vector missing

These are prevented by the json schema of cvss v3.x, but the validation against that is only done in 6.1.8.
To this test, the cvss_v3 object is a plain json object, which could contain anything.

-> Opened an issue for this: #578

### DIFF
--- a/csaf-rs/src/csaf2_0/validation.rs
+++ b/csaf-rs/src/csaf2_0/validation.rs
@@ -166,7 +166,7 @@ impl Validatable for CommonSecurityAdvisoryFramework {
             match test_id {
                 // informative tests
                 "6.3.1" => Some(ValidatorForTest6_3_1.validate(self)),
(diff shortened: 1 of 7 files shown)


tziemek (approved):

peinjoh:
Clippy got updated upstream, including additions to the default ruleset, had two fix two files :)

PR #612, opened 2026-04-30
Fix no-default-features build.

SimonHeidrich:
The tsify dependency is (correctly) marked as optional, because it is only required by the "wasm" feature. However, it is unconditionally used in validation.rs. Since the "wasm" feature flag is enabled by default, this only becomes relevant if default features are disabled.

This PR adds a verification test for the build without default features and guards the usage of tsify behind a feature flag.


tziemek:
Thank you for your pull request. We can put that in. 
It would have been no issue in the near future, as we have a PR in draft , where the wasm feature is replaced by UniFFI generation, but until this is complete, this is a good change and we keep the change to the build pipeline to catch these kind of things in the future.

tziemek (approved):

Issue #625, opened 2026-05-05
Improve 6.1.7 test coverage


peinjoh:
TODO:
* backport 2.1 tests to 2.0 where applicable
* non-cvss metrics should have no effect

PR #626, opened 2026-05-05
feat: revisit test 6.1.7 for CSAF 2.1

### LINKED ISSUE #341: Implement 6.1.7 Multiple Scores with Same Version per Product for CSAF2.1

peinjoh:
This PR revisits / cleans up the test 6.1.7 impl for CSAF 2.1.

This includes:
* extracting the product -> cvss metrics aggregation
* improved documentation of test cases
* improve code comments and code cleanup

I found some room for improvement in the test coverage: #625 

Resolves #341


aschierl-xitaso:
At the moment I don't see why we want to extract this specific aggregation and make it reusable (other users might well need them as "product id > source > metric type" or "source > product id"). I could imagine extracting a collection of struct { product_id, source, metric_type, path} that can be used more flexibly.

If we just need the uniqueness of the three values, we could also put those three as a tuple into the HashMap key instead of using nested HashMaps.

tziemek (commented):
I'm with @aschierl-xitaso here no need to extract, at least not as `pub`.

peinjoh:
> At the moment I don't see why we want to extract this specific aggregation and make it reusable (other users might well need them as "product id > source > metric type" or "source > product id"). I could imagine extracting a collection of struct { product_id, source, metric_type, path} that can be used more flexibly.
> 
> If we just need the uniqueness of the three values, we could also put those three as a tuple into the HashMap key instead of using nested HashMaps.

As discussed, this will be done in a seperate PR which introduces a generic helper function (Issue: #649)

peinjoh:
> I'm with @aschierl-xitaso here no need to extract, at least not as `pub`.

Reverted :)

tziemek (approved):

PR #635, opened 2026-05-15
fix: run apple x86_64 build on intel

aschierl-xitaso:


aschierl-xitaso:
fixed in #637 using cross compilation

PR #636, opened 2026-05-15
build: update github actions to latest versions

tziemek:


.github/workflows/build.yml
```diff
@@ -499,10 +525,11 @@ jobs:
         uses: actions/setup-go@v5
```
peinjoh: Minor: this could also be v6, right?

peinjoh (approved):
LGTM 🪨

PR #638, opened 2026-05-17
fix: create validators on demand for 6.1.8

tziemek:
Instead of creating all 4 validators on every function call, we create them on demand depending on the validated document. This should save a lot of time when those validator aren't even needed.


peinjoh (approved):
LGTM 🪨 

We might also want to move these lazys into a helper (the other lazy validators needed)?

PR #639, opened 2026-05-17
feat: add load_from_value

tziemek:
Adds a function to load the document from an already parsed serde_json:Value. This can prevent parsing the document twice.


peinjoh (approved):
LGTM 🪨

PR #640, opened 2026-05-17
feat: adds validate _as_json functions for WASM

tziemek:
This adds validate functions to UniFFI exports which return the result as a String. In WASM the JSON.from is highly performant and can handle a raw string much better than a typed value, because of how UniFFI is creating an interpreted typescript parser for it's binary protocol.

### DIFF
--- a/csaf-ffi/src/lib.rs
+++ b/csaf-ffi/src/lib.rs
@@ -150,7 +150,7 @@ impl From<csaf::validation::ValidationResult> for ValidationResult {
 /// # Arguments
 ///
 /// * `json_str` - The CSAF document as a JSON string.
-/// * `preset`   - The validation preset: `"basic"`, `"extended"`, or `"full"`.
+/// * `preset`   - The validation preset: e.g. `"basic"`, `"extended"`, or `"full"`.
 ///
 /// # Returns
 ///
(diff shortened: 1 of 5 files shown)


csaf-ffi/src/lib.rs
```diff
@@ -211,3 +211,81 @@ pub fn validate_csaf_2_1(json_str: String, preset: String) -> Result<ValidationR
...
+///
+/// Same as [`validate_csaf`] but returns the [`ValidationResult`] serialized
+/// as a JSON string instead of a typed UniFFI record.  Prefer this in WASM
+/// contexts: `JSON.parse(result)` is much faster than the UniFFI binary
+/// deserializer for the result type.
+///
+/// # Arguments
+///
+/// * `json_str` - The CSAF document as a JSON string.
+/// * `preset`   - The validation preset: `"basic"`, `"extended"`, or `"full"`.
```
peinjoh: Minor: Add an e.g. here to highlight that these three are not the only presets available (on CSAF 2.1)

peinjoh (approved):
Side note: This PR adds a bit of code duplication. Fine for now, but if more functions are added in the future, we might want to extract the shared code.

Issue #205, opened 2025-12-05
Consider changing types of simple attributes to non-ref


tziemek:
Basically all return types of the traits are returning a ref, which leads to a lot of noise with `&`, `deref` etc. This shouldn't be necessary, especially for simple types as copying them is ok as it happens and the stack.

PR #657, opened 2026-05-20
refactor: simplify trait return types

peinjoh:
Relates to #205 

This PR: 
* moves &String to &str
* puts Copy on some small enums / types and uses it in the codebase

This improves readability and minimally improves performance for the types that are more efficiently passed by value.

### DIFF
--- a/csaf-rs/src/validations/test_6_2_01.rs
+++ b/csaf-rs/src/validations/test_6_2_01.rs
@@ -25,7 +25,7 @@ pub fn test_6_2_01_unused_definition_of_product_id(doc: &impl CsafTrait) -> Resu
     }
 
     // Get all references to product IDs in the document
-    let references = doc.get_all_product_references_ids();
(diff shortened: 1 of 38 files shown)


csaf-rs/src/validations/test_6_2_01.rs
```diff
@@ -25,7 +25,7 @@ pub fn test_6_2_01_unused_definition_of_product_id(doc: &impl CsafTrait) -> Resu
     }
 
     // Get all references to product IDs in the document
-    let references = doc.get_all_product_references_ids();
```
peinjoh, line 28: Only functional change: During refactoring, I realized that that the references can be deduplicated, as they are only used in the contains check below.

tziemek (approved):
Thanks, so much easier handling, especially with the enums.

peinjoh:
Performance fluctuations are in range of what is to be expected by github runner performance fluctuation. Fine as is, merging 🚀

PR #661, opened 2026-05-21
refactor: `generic_with` impls via macro

peinjoh:
Did a little experiment on how our traits could look if we leaned more into macros for the repetitive code. I look forward to your thoughts :)

Pros: brevity, DRY, improved readibility for experienced devs
Cons: worse readability for first contributors, debugging can get wierd on some IDEs

Please look over the lack of polish, this is meant as a demo :)

### DIFF
--- a/csaf-rs/src/csaf/traits/util/generic_with.rs
+++ b/csaf-rs/src/csaf/traits/util/generic_with.rs
@@ -5,66 +5,55 @@ pub trait WithOptionalGroupIds {
     fn get_group_ids(&self) -> Option<impl Iterator<Item = &String> + '_>;
 }
 
-/// Implements `WithOptionalGroupIds` by delegating to `self.group_ids`.
-macro_rules! impl_with_optional_group_ids {
-    ($type:ty) => {
-        impl $crate::csaf_traits::WithOptionalGroupIds for $type {
-            fn get_group_ids(&self) -> Option<impl Iterator<Item = &String> + '_> {
-                self.group_ids
-                    .as_ref()
-                    .map(|g| g.iter().map(|x| ::std::ops::Deref::deref(x)))
-            }
-        }
-    };
-}
-
-pub(crate) use impl_with_optional_group_ids;
-
-/// Implements `WithOptionalGroupIds` as always returning `None`.
-macro_rules! impl_without_group_ids {
-    ($type:ty) => {
-        impl $crate::csaf_traits::WithOptionalGroupIds for $type {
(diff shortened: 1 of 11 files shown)


csaf-rs/src/csaf/traits/util/generic_with.rs
tziemek, line 1: It feels like the `with` and `without` macros could be combined with a parameter.

tziemek (commented):
Hmm, I don't know about that. We can start with these simple ones which really make the code more concise.

csaf-rs/src/csaf/traits/util/generic_with.rs
peinjoh: Sure can, you mean returning &str and Option<&str>?

tziemek: Sorry that was confusing. I meant providing a parameter to choose whether to return the value or None.

peinjoh: I can see that we want to de-duplicate the code, but I would prefer to not introduce some boolean flag to decide if the actual impl or none is returned.

tziemek: Fine for me now, maybe consider this when we have many more of those macros. Also i'm not happy with the naming, but have no better suggestion at the moment, so keep it.

peinjoh: Merged into one macro that does WithOptionalProductId / GroupId / ReturnsValues / ReturnsEmpty.

I considered doing it with macro branching do DRY the 4 branches, but I would argue thats overkill for this specific case.

tziemek: Oh that looks nice

tziemek (approved):

csaf-rs/src/csaf/traits/util/generic_with.rs
```diff
@@ -10,12 +10,85 @@ pub trait WithOptionalProductIds {
     fn get_product_ids(&self) -> Option<impl Iterator<Item = &String> + '_>;
 }
 
+/// Implements `WithOptionalGroupIds` or `WithOptionalProductIds` for a type.
+///
+/// If `ReturnsValues` is specified, the implementation expects `group_ids` or `product_ids`
+/// to be an `Option<Vec<String>>` field in the struct.
+///
```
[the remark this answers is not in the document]
peinjoh: Will update this in the next PR.

PR #675, opened 2026-06-05
feat: abstract the source of the JSON content

ctron:
Instead of having multiple variants of "load_abc" functions, create a trait and let it implement what "loading JSON" means. And have the content functions just re-use that.

### DIFF
--- a/csaf-rs/src/csaf2_0/loader.rs
+++ b/csaf-rs/src/csaf2_0/loader.rs
@@ -1,27 +1,7 @@
-use crate::{csaf::raw::RawDocument, schema::csaf2_0::schema::CommonSecurityAdvisoryFramework};
-use std::{fs::File, io::BufReader};
+use crate::{csaf::raw::RawDocument, json::JsonSource, schema::csaf2_0::schema::CommonSecurityAdvisoryFramework};
 
-pub fn load_document(path: &str) -> std::io::Result<RawDocument<CommonSecurityAdvisoryFramework>> {
-    let f = File::open(path)?;
-    let reader = BufReader::new(f);
-    let doc: RawDocument<CommonSecurityAdvisoryFramework> = RawDocument::new(serde_json::from_reader(reader)?);
-    Ok(doc)
-}
-
-/// Load a CSAF document from a JSON string
-pub fn load_document_from_str(json_str: &str) -> std::io::Result<RawDocument<CommonSecurityAdvisoryFramework>> {
-    let doc: RawDocument<CommonSecurityAdvisoryFramework> = RawDocument::new(serde_json::from_str(json_str)?);
-    Ok(doc)
-}
-
-/// Load a CSAF document from an already-parsed [`serde_json::Value`].
-///
-/// Use this when the JSON has already been parsed (e.g. to extract the
-/// `document.csaf_version` field) to avoid parsing the same string twice.
-pub fn load_document_from_value(
-    value: serde_json::Value,
-) -> Result<RawDocument<CommonSecurityAdvisoryFramework>, serde_json::Error> {
-    Ok(RawDocument::new(serde_json::from_value(value)?))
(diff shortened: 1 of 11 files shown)


csaf-rs/src/csaf2_0/loader.rs
```diff
@@ -1,27 +1,24 @@
...
-    let f = File::open(path)?;
-    let reader = BufReader::new(f);
-    let doc: RawDocument<CommonSecurityAdvisoryFramework> = RawDocument::new(serde_json::from_reader(reader)?);
-    Ok(doc)
+pub fn load_document<T: JsonSource>(source: T) -> std::io::Result<RawDocument<CommonSecurityAdvisoryFramework>> {
+    Ok(RawDocument::new(source.parse()?))
 }
 
 /// Load a CSAF document from a JSON string
+#[deprecated = "Use `load_document` directly"]
```
tziemek: No need for deprecation at the moment, as we don't have a stable API. You can just delete it.

csaf-rs/src/csaf2_1/loader.rs
```diff
@@ -1,27 +1,24 @@
...
-    let f = File::open(path)?;
-    let reader = BufReader::new(f);
-    let doc: RawDocument<CommonSecurityAdvisoryFramework> = RawDocument::new(serde_json::from_reader(reader)?);
-    Ok(doc)
+pub fn load_document<T: JsonSource>(source: T) -> std::io::Result<RawDocument<CommonSecurityAdvisoryFramework>> {
+    Ok(RawDocument::new(source.parse()?))
 }
 
 /// Load a CSAF document from a JSON string
+#[deprecated = "Use `load_document` directly"]
```
tziemek: No deprecation needed, just delete it.

csaf-rs/src/json.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Test from a path to a file
+    #[test]
+    fn from_path() {
+        let t = tempfile::tempdir().expect("must create temp dir");
+
+        let path = t.path().join("test.json");
+        fs::write(&path, br#"{}"#).expect("must write to temp file");
+
+        let _ = load(&path).expect("must load");
+        let _ = load(&path as &Path).expect("must load");
```
tziemek, line 161: Maybe close the directory manually to make sure it is really gone.
```suggestion
        let _ = load(&path as &Path).expect("must load");
        t.close().expect("must delete temp dir");
```

tziemek (changes requested):
Thanks for your pull requests, I really like these changes, but have a few comments.
Please pull in the latests main branch, the pipeline should be fixed now.

csaf-rs/src/json.rs
ctron: To my understanding, this is done in `drop` anyway. Having an explicit close would only raise that error. Made the change.

csaf-rs/src/csaf2_1/loader.rs
ctron: Done

csaf-rs/src/csaf2_0/loader.rs
ctron: Done

ctron:
I think I made all requested changes, unless I missed something. And did a re-base.

csaf-rs/src/json.rs
tziemek: Yes, it should be handled by the `drop`, but the docs explicitly say a manual close is recommended
```
// By closing the `TempDir` explicitly, we can check that it has
// been deleted successfully. If we don't close it explicitly,
// the directory will still be deleted when `dir` goes out
// of scope, but we won't know whether deleting the directory
// succeeded.
dir.close()?;
```

tziemek (approved):

PR #676, opened 2026-06-05
feat: generate/check result json

aschierl-xitaso:
This spike extends the validator cli by adding options to create a test result file for a given input file and test, or to check a result file against the test results for a given input file. It currently uses a modified copy of the result file schema, and does not support result files in testcases.json, and is meant for experimenting with result files.


tschmidtb51:
> It currently uses a modified copy of the result file schema, and does not support result files in testcases.json

Is the only modification the resolution of the pattern?

> does not support result files in testcases.json

Not in scope for the first draft.

The result.json should be sorted.
The docs should clarify, what the relation between `-T` and `--create-test-result` is.

aschierl-xitaso:
> Is the only modification the resolution of the pattern?

Yes, the only modification is the pattern description, because the `typify` based type generation does not support referencing external schemas.

> The result.json should be sorted.

Good catch, seems like `typify` sorts the fields by letters, ignoring the leading dollar sign. We will also have to treat this when outputting CSAF documents, also there `$schema` is placed incorrectly.

tziemek:
See https://github.com/oasis-tcs/csaf/issues/820 for decisions.

tziemek (approved):
LGTM

PR #682, opened 2026-06-15
build: fix SBOM upload

tziemek:


bump_version.sh
```diff
@@ -0,0 +1,88 @@
...
+
+echo "Bumping version: $CURRENT_VERSION -> $NEW_VERSION"
+
+# --- Update files ----------------------------------------------------------
+
+# workspace Cargo.toml: [workspace.package] version field
+# workspace Cargo.toml: the [workspace.package] version field (first occurrence)
+sed -i.bak "s/^version = \"[^\"]*\"/version = \"${NEW_VERSION}\"/" Cargo.toml
+
+# path dependencies on csaf-rs in csaf-validator and csaf-converter
```
peinjoh: Would it make sense to generalize this more, so we don't need to update this list if more tools are added?

peinjoh (approved):

bump_version.sh
tziemek: Now using cargo-edit and npm instead of sed to bump versions

PR #684, opened 2026-06-15
fix: use Path instead of string content for parsing

### LINKED ISSUE #683: Creation of test result JSONs fails
After the [latest merge in main](https://github.com/csaf-rs/csaf/pull/676), the function seem to fail. I use the command:

``` 
./csaf-result-json create 6.1.1 ~/path-to-oasis-repo/csaf/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-01-01.json
```

but get the result:
```
Error: expected value at line 1 column 1
```

tziemek:
resolves #683 

The detect_version function was called with the name of the file instead of the content of the file. This was probably introduced with the `JsonSource` functionality. This is a quick fix which uses the file path, so the content passed to the functions is really the json content.

### DIFF
--- a/csaf-result-json/src/main.rs
+++ b/csaf-result-json/src/main.rs
@@ -77,7 +77,6 @@ fn main() -> Result<(), anyhow::Error> {
             csaf_document: path,
         } => {
             tests.push(test_id.as_str());
-            println!("Running tests: {tests:?}");
 
             let file_path = Path::new(&path);
 
@@ -85,7 +84,6 @@ fn main() -> Result<(), anyhow::Error> {
                 "auto" => detect_version(file_path)?,
                 other => other.to_string(),
             };
-            println!("Detected CSAF version: {version}");
 
             let result = match version.as_str() {
                 "2.0" => {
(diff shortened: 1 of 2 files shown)


csaf-result-json/src/main.rs
```diff
@@ -75,19 +77,23 @@ fn main() -> Result<(), anyhow::Error> {
             csaf_document: path,
         } => {
             tests.push(test_id.as_str());
+            println!("Running tests: {tests:?}");
```
aschierl-xitaso: I wouldn't put `println!` in the default output path, this breaks some `csaf-result-json create ... > result.json` workflows.
Either use stderr / eprintln, or leave them out completely.

peinjoh:
Could you add a comment explaining what the bug is and how this changeset fixes it?

tziemek:
> Could you add a comment explaining what the bug is and how this changeset fixes it?

Edited the description.

aschierl-xitaso (approved):

PR #691, opened 2026-06-22
Fix 6.2.5

georgbramm:
Fixed test code for test 6.2.5


aschierl-xitaso:
Sorry, I don't really understand - why would you expect an error message
```
 /document/tracking/initial_release_date: Initial release date '2023-09-06T10:00:00.000Z' is older than the earliest revision history date '2023-08-22T10:00:00.000Z' [Warning 6.2.5]
⚠️  Validation passed with 1 warning(s) and 0 info(s).
```
in test file `oasis_csaf_tc-csaf_2_1-2024-6-2-05-01.json`
```
    "tracking": {
      "current_release_date": "2024-01-21T10:00:00.000Z",
      "id": "OASIS_CSAF_TC-CSAF_2.1-2024-6-2-05-01",
      "initial_release_date": "2023-08-22T10:00:00.000Z",
      "revision_history": [
        {
          "date": "2023-09-06T10:00:00.000Z",
          "number": "1",
          "summary": "Initial version."
        },
        {
          "date": "2024-01-21T11:00:00.000Z",
          "number": "2",
          "summary": "Second version."
        }
      ],
      "status": "final",
      "version": "2"
    }
```
that obviously does not have an initial release date of `2023-09-06T10:00:00.000Z`?

georgbramm:
You are absolutely right!
i fixed the code by finding the true oldest revision by iterating and finding the minimum UTC timestamp

georgbramm:
or should i use rev_history.inplace_sort_by_date_then_number()?

georgbramm:
I adopted it to use the provided function "inplace_sort_by_date_then_number()"

tschmidtb51:
> You are absolutely right! The issue stems from an upstream bug in the OASIS CSAF 2.0 test cases, where the expected error string has inverted timestamps compared to the actual JSON test asset. I mistakenly inverted the arguments in the core validation logic to satisfy the broken 2.0 test runner, which broke the output for CSAF 2.1. I will refactor it to keep the validation logic clean and strictly handle the 2.0 discrepancy inside the unit test wrapper only. I will push the fix in the next few minutes.

I'm confused. Could you please detail where the "upstream bug" is/was? I looked at https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/test/validator/data/optional/oasis_csaf_tc-csaf_2_0-2021-6-2-05-01.json and https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#625-older-initial-release-date-than-revision-history - but I'm not seeing it atm.

georgbramm:
Hi!

You are completely right, and I owe you a big apology—there is no upstream bug in the OASIS test cases or the specification. The confusion was entirely on my end due to a logical knot I tied myself into while debugging. I mistakenly flipped the comparison operator after misidentifying a valid CSAF 2.0 test file as a failing one, which corrupted the timestamp ordering and broke the CSAF 2.1 validation output.

The Fix:
I have refactored the implementation to strictly use the mathematically correct < operator as defined by the specification. Both CSAF 2.0 and 2.1 test suites are running fully green locally now. I’ve just pushed the clean fix.

Thanks a lot for double-checking and calling this out—it completely put me back on the right track!

aschierl-xitaso:
If I see it correctly, the pull request

- includes the unrelated change of #690
- changes comments, and renames variables in test 6.2.5
- introduces white space in the middle of a word in test 6.2.6

but contains no substantial contribution or change in logic.

georgbramm:
yes, you see that right. 
i finally ended up in the same code as it was before. (thats also a reason why i initially thought there must be something wrong with the test cases)
It seems like there was some kind of miscommunication between csaf and AISEC.
we were told to implement the following test: 6.3.11, 6.2.3, 6.2.4, 6.2.5, 6.2.6

but it seems that these tests were already implemented (besides the None->Some change in 6.3.11), so we were a bit confused.
We just informed @oxisto about that and are currently waiting for feedback.....

tziemek:
You surely can implement these tests. We have a lot covered for CSAF 2.0, but still missing the 2.1 implementation. Meaning all issues labeled with csaf 2.1 or having "test for csaf 2.1" in their title and have a corresponding test in CSAF 2.0 may already have an implementation which works for CSAF 2.0 files. You have to check if it also works for CSAF 2.1 files and possibly change the implementation so it works for both versions. Please also update the validation.rs file and the README to reflect the implementation status. If necessary, add additional supplementary test cases (may alread be there for 2.0, missing for 2.1).

Issue #571, opened 2026-04-14
Allow lenient unparsing for edit mode


peinjoh:

peinjoh:
Outlook on how we could implement this:

Currently, all tests run on `&impl CsafTrait`, which is implemented by the CSAF 2.0 and CSAF 2.1 implementations, which use the typify-generated schema implementations.

We propose to "split" the test implementation into a "pre-validation" and a "core validation", with the core validation receiving custom structs that only contain the "relevant" test data. The pre-validation contain the access to the relevant data via JSONPath, instead of via traversal of the traits in the current implementation. 

Example for test 6.1.12:
For this test,  `/document/lang` and `/document/source_lang` are relevant, with the core test logic being the validation of the language tags.

* In the "pre-validation" implementation, we would extract the values of the two fields via JSONPath.
* The "core validation" implementation, we would receive a struct with two Strings, which we would then a) parse into the typgen type to apply the regex validation of the json schema and b) validate via our custom implementation `CsafLanguage`, which checks against the IANA registry.

Example for test 6.2.12:
For test 6.2.12, which requires only `/document/lang` with the core test logic being that it needs to be present and set.

* In the "pre-validation" implementation, we would extract the document language via JSON Path. If the key is not present, we return an error here.
* In the "core validation" implementation, we receive the present string, and check if the string is empty.

Example for test 6.2.14:
For test 6.2.14, `/document/lang` and `/document/source_lang` are relevant, with the core test logic being to check that no private tags are used.

* In the "pre-validation" implementation, we would extract the values via JSON Path, apply the regex from the schema via the typegen types and apply our custom parsing and validation.

* In the "core validation" implementation, we receive a struct with two `Option<CsafLanguage>`, for which we can check that in case of the `Option`s being `Some` (i.e. the properties were present, set, passed regex, passed parsing), if the parsed language tags are Private Use.

This will implement the "lenient" parsing required to handle documents that do not pass the schema validation. For documents that are modified / generated by our libary, we can then use the previous "strict" parsing via compliance with the CsafTraits implementations.

PR #693, opened 2026-06-22
feat: add extractor-based, non-strict parsing

aschierl-xitaso:
CSAF validators should be able to work on partially invalid documents (according to the schema) and allow performing tests as long as the invalid parts do not affect the test logic. For this, we intend to implement non-strict parsing and validation. 

This pull request proposes two alternatives to achieve this goal: 
- One approach that splits tests into two parts, one that extracts all relevant data for the test from the JSON document in a stream-based fashion while iterating the JSON just once, and a second part that performs the validation on the extracted data. This approach allows to parse streaming data (and syntactically invalid JSON up to the first error) and can avoid deserializing the whole document into memory, but causes more changes to the structure of tests and cannot use the existing abstractions of the traits. 
- One approach that makes the existing traits non-strict (i.e. every field is an Option<T> that can be None if the JSON does not contain a valid value for that), and implements the traits upon reference wrappers around `serde_json::Value`. This leaves the test software structure similar to the current state and just requires the tests to with the newly added None options.

For more details, see the ADR document in the pull request. 

For the rough idea, see #571, although some details have been implemented differently)


gronke:
A data point in favor of this direction, from consuming the generated models for document intake:

Because the schema patterns are enforced through `regress` inside `Deserialize`, one cosmetically nonconforming field (say a `cve` failing the CVE pattern) rejects the entire document — there is no way to parse the 99% that is fine and inspect the rest. And since the typed model cannot retain the verbatim JSON, an ingest that needs both ends up parsing twice: `from_str` to `Value`, clone, then `from_value` into the typed document. An extractor-based, non-strict path would serve both: read what parses, report what does not, and leave the raw `Value` available alongside.

aschierl-xitaso:
To get an indication about the performance possible with the different approaches, I tried parsing a 32MB CSAF document using different approaches: 
- Stream-based extraction the document category (see test 6.2.38)
- Stream-based parsing into serde_json::Value (as benchmark against direct serde_json deserialization)
- Parsing the document into serde_json::Value (for use with Value-based traits)
- Parsing the document into serde_json::Value, then CommonSecurityAdvisoryFramework as generated by typify
- Parsing the document directly into CommonSecurityAdvisoryFramework as generated by typify

| Approach  | Duration | Memory used  |
| ------------- | ------------- | -------------- |
| Stream-based extraction | 0.07s  | 7MB |
| Parsing into CSAF | 0.13s | 61MB |
| Parsing into serde_json::Value  | 0.16s  | 94MB |
| Stream-based into serde_json::Value | 0.16s | 107MB |
| Parsing into Value, then CSAF | 0.16s | 113MB |

Especially in cases where only selected tests run and not all data is required, stream-based parsing has the potential to improve speed and memory usage (memory grows only with the size of extracted data, not linear with the size of the whole input file). 
Using serde_json::Value as an intermediate format increses the memory usage and run time, so directly deserializing into a non-strict object model of CSAF might also be an efficient solution (not considered in this pull request).

Issue #159, opened 2025-11-21
CSAF 2.0 docs testable "as CSAF 2.1 docs"


milux:
> TLP is checked in a recommended test in v2.0, which is obsolete in v2.1 because it is directly in the schema. So I think we should have `Result` in both cases.

I see, right. However, I would still keep it as that, because for CSAF 2.0, it is an optional element, and whenever I did this `_20` and `_21` things, I wanted to explicitly denote different semantics for different CSAF versions.

If you still think that this is not helpful (I'm fine with that), go ahead, drop the `_20` variants and the `_21` suffix and make it one function returning `Result`.

_Originally posted by @milux in https://github.com/csaf-rs/csaf/pull/155#discussion_r2549072622_

milux:
@tziemek @tschmidtb51 When writing that, it came to my mind that I made a design decision that not everyone must necessarily agree with.

In the way I wrote the tests, a user may test a CSAF 2.0 document as if it was a CSAF 2.1 document. The results of such validations are - naturally - not well defined in every chase.
If fields do not exist, such validations may fail, but they may also return `Ok` if there are optional semantics in CSAF 2.1 and the fields under test simply don't exist.

Just wanted you to know about that and keep it in mind. One may instead block the execution of 2.0 tests on 2.1 docs entirely. That's up to you now. :)

PR #695, opened 2026-06-23
Implementing 6.2.23 Usage of Deprecated CWE

### LINKED ISSUE #309: Implement 6.2.23 Usage of Deprecated CWE for CSAF 2.1

konradweiss:
This PR:

- Adds `Status` extraction to the CWE download and updates the CWE files.
- Implements the check for status==deprecated to fail validation.
- No further testcases.

Resolves #309


konradweiss:
Closes #309

aschierl-xitaso:
I think, looking for `DEPRECATED:` in the title is not sufficient, for example CWE 1.10 contains `CWE 218` with title `DEPRECATED (Duplicate): Failure to provide confidentiality for stored data`.

Feel free to extend the extraction process (`scripts/update/update_all_cwe.sh` / `scripts/update/convert-cwe-to-csv.xslt`) to also extract further required fields, e.g. the `Status` field. In the future, once we have collected the data needed for all test, we will probably move the data and extraction process to https://github.com/csaf-rs/cwe and also add a rust crate providing the data - but for now, having the required data as CSV files (and thus knowing what data is needed) should be sufficient.

konradweiss:
> I think, looking for `DEPRECATED:` in the title is not sufficient, for example CWE 1.10 contains `CWE 218` with title `DEPRECATED (Duplicate): Failure to provide confidentiality for stored data`.
> 
> Feel free to extend the extraction process (`scripts/update/update_all_cwe.sh` / `scripts/update/convert-cwe-to-csv.xslt`) to also extract further required fields, e.g. the `Status` field. In the future, once we have collected the data needed for all test, we will probably move the data and extraction process to https://github.com/csaf-rs/cwe and also add a rust crate providing the data - but for now, having the required data as CSV files (and thus knowing what data is needed) should be sufficient.

Thanks, I will further look into that

konradweiss:
I would consider this PR to be finished now, I have not added any new test files, as the existing ones appear to cover the relevant cases. Don't be shocked about the number of changed files. The csv files containing CWE data were outdated after me adding the status extraction and are therefore updated.

aschierl-xitaso:
Feel free to also include CWE version 4.20, which has been released since the last update of our CWE list.
(Unfortunately, new CWE releases are not yet covered by our automatic asset update process).
Just add `v4.20` to `scripts/update/update_all_cwe.sh` and re-run the script.

csaf-rs/src/validations/test_6_2_23.rs
```diff
@@ -0,0 +1,83 @@
+use crate::csaf_traits::{CsafTrait, VulnerabilityTrait};
+use crate::helpers::CWE_ENTRIES;
+use crate::validation::ValidationError;
+
+fn create_deprecated_cwe_error(cwe: &str, version: &str, path: &str) -> ValidationError {
+    ValidationError {
+        message: format!("Weakness '{cwe}' is deprecated in version {version}."),
+        instance_path: format!("{path}/name"),
```
aschierl-xitaso: I think the `name` of the CWE is no longer the relevant element of the error (rather the `id` or the CWE itself)

konradweiss: I changed instance_path to "{path}/id" is that sufficient? I hope I understood the issue correctly.

aschierl-xitaso: The instance path is used to show what is wrong (e.g. highlighted in SecVisoGram), or what should be changed to fix the document. 

In this case, I'm not fully sure what is the best option, it could be:
- change `version` back to a version where this was not yet deprecated - then we would report `/vulnerabilities/0/cwes/0/version` as instance path (but that older version would trigger test `6.2.24 Usage of Non-Latest CWE Version`)
- change `id` to choose a non-deprecated CWE - then it's `/vulnerabilities/0/cwes/0/id` (but this will trigger test `6.1.11 CWE` when the `name` is not changed correspondingly)
- change the whole CWE object to a new `{id, name, version}` tuple - then it's `/vulnerabilities/0/cwes/0` (that's also what I think the standard hint "A tool MAY suggest to replace the deprecated CWE with its replacement or closest equivalent." implies)

I slightly prefer the third, but can live with the second.

tziemek: Secvisogram currently uses the third option.

csaf-rs/src/validations/test_6_2_23.rs
```diff
@@ -0,0 +1,83 @@
...
+            for (i_cwe, cwe_item) in cwes.iter().enumerate() {
+                // Determine which CWE CSV version to check: use the version
+                // declared on the CWE item
+                let Some(version) = cwe_item.version.as_deref() else {
+                    // version missing — Document should already have been rejected
+                    continue;
+                };
+
+                // We can also just run this part, and only look into the cwe file
+                if let Some((status, _)) = CWE_ENTRIES[version].1.get(&cwe_item.id)
```
peinjoh, line 34: There is a missing precondition check here: 6.1.11 checks if the version is one of the known versions first (via `contains()`). This check is missing here I think. Please check how your impl acts on cwe-version-shaped, but currently non-existent cwe versions (like 9.9). Currently, I believe this will panic. Please add test coverage to test this behaviour, or consider adding a wrapper around CWE_ENTRIES that encompasses this check.

csaf-rs/src/validations/mod.rs
```diff
@@ -118,6 +118,7 @@ pub mod test_6_3_05;
 // pub mod test_6_3_06;
 // pub mod test_6_3_07;
 // pub mod test_6_3_08;
+pub mod test_6_2_23;
```
peinjoh: Minor: This file is lexicographically ordered, please move this line appropriatly.

peinjoh (changes requested):
Final though: We have #159 in the backlog, and this test would be a candidate to implement this.

In test 6.1.11 for 2.0, we derive the cwe version from the `current_release_date` field. We could do the same in this test (which did not exist in 2.0) to run the same check for deprecated cwes, based on the derived cwe version. This does not need to be done in this PR.

EDIT: I want to underline that this is explicitly not part of this PR and not part of the expected problem/issue resolution. See below: https://github.com/csaf-rs/csaf/pull/695#issuecomment-5034385674

oxisto:
> Final though: We have #159 in the backlog, and this test would be a candidate to implement this.
> 
> In test 6.1.11 for 2.0, we derive the cwe version from the `current_release_date` field. We could do the same in this test (which did not exist in 2.0) to run the same check for deprecated cwes, based on the derived cwe version. This does not need to be done in this PR.

I'd strongly suggest that we keep the scope of the original ticket, which explicitly states "CSAF 2.1" in the name. Otherwise we might offload tasks to external contributors which they would not foresee when originally looking at a ticket to implement. Better to make a dedicated new ticket for this and if there is interest in contributing to that, they surely will.

peinjoh:
> I'd strongly suggest that we keep the scope of the original ticket, which explicitly states "CSAF 2.1" in the name. Otherwise we might offload tasks to external contributors which they would not foresee when originally looking at a ticket to implement. Better to make a dedicated new ticket for this and if there is interest in contributing to that, they surely will.

I completly agree, and would underline that the "final though" was explicitly not meant as a "you should do that too", but as somewhere between "if we consider #159, this might be relevant" and "if someone wants to experiment, this is a starting point" and definitily as "this is outside of the scope of this PR". I see that this could have been communicated clearer 👍

csaf-rs/src/validations/test_6_2_23.rs
konradweiss: I updated the path to option 3 and simplified the error creation function to directly get the vulnerabilities and cwe index.

csaf-rs/src/validations/test_6_2_23.rs
konradweiss: I provided a check that will skip the deprecation check for the cwe if the cwe version does not exist. This should be fine if an error is already generated when 6.1.11 encounters the non existing CWE version.

csaf-rs/src/validations/mod.rs
konradweiss: Done

csaf-rs/src/validations/test_6_2_23.rs
```diff
@@ -0,0 +1,91 @@
...
+                "4.13",
+                2,
+                0,
+            )]),
+            // case 11: valid
+            Ok(()),
+            // case 12: valid
+            Ok(()),
+            // case 13: valid
+            Ok(()),
```
peinjoh: Minor, as discussed in person: Please extract the errors into speaking-named variables, and add one-liners on what the passing cases entail.

csaf-rs/src/validations/test_6_2_23.rs
peinjoh: Looking good!

csaf-rs/src/validations/test_6_2_23.rs
peinjoh: Looks good to me, I think this is resolved, and I agree that option 3 aligns best with our other implementations. @aschierl-xitaso @tziemek please reopen if you see more work here.

csaf-rs/src/validations/test_6_2_23.rs
peinjoh: Looks fine 👍

peinjoh (approved):
LGTM, is there something left to do here @konradweiss ?

konradweiss:
> LGTM, is there something left to do here @konradweiss ?

No, this IMO addresses all the open issues and is now finished.

peinjoh:
> > LGTM, is there something left to do here @konradweiss ?
> 
> No, this IMO addresses all the open issues and is now finished.

Alright, merging 👍

peinjoh:
Thanks again for your contribution 🚀

PR #696, opened 2026-06-23
feat: implement validation test 6.2.31 for hardware and software path…

### LINKED ISSUE #317: Implement 6.2.31 Hardware and Software for CSAF 2.1

### LINKED ISSUE #701: Wrong `instance_path` reported for `product_paths`
In the `product_tree_trait.rs` we report `/product_tree/relations/...` for full product names in either csaf 2.0 and 2.1 as it is gathered in the `...generic` function.
```rust

        // Visit relationships
        for (i, rel) in self.get_product_paths().iter().enumerate() {
            callback(
                rel.get_full_product_name(),
                &format!("/product_tree/relationships/{i}/full_product_name"),
            );
        }
```
This reports `relations` for a CSAF 2.1 document, where this is moved to `product_paths`.
We should move this to the version dependent implementations and also check if the `instance_path` is correct (not only the relationships/product_paths node.

georgbramm:
Implemented the mandatory CSAF Test 6.2.31 (Hardware and Software) validation logic to detect invalid mixtures of hardware and software products within the product tree.

Changes

- Added src/validations/test_6_2_31.rs:
- Aggregated all valid target product identifiers from product_tree.get_product_paths() by extracting the beginning references, full product IDs, and any relationship references via .get_subpath_product_references().
- Utilized the framework's native visit_all_products_generic visitor pattern to recursively traverse all products across nested branches without manual recursion loops.
- Verified if a product defines serial_numbers or model_numbers via the ProductIdentificationHelperTrait getters (.get_serial_numbers() / .get_model_numbers()).
- Enforced that an error is generated if a hardware identifier lacks a corresponding structural product path.
- Updated src/validations/mod.rs: Registered the new validation module into the validation pipeline execution.

Verification
Ran cargo test to verify compliance. All underlying framework validation test cases (case_01 through case_13) pass successfully.


csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
+use crate::csaf::traits::vulnerabilities::product_ident_helper_trait::ProductIdentificationHelperTrait;
+use crate::csaf_traits::{CsafTrait, ProductPathTrait, ProductTrait, ProductTreeTrait};
+use crate::validation::ValidationError;
+
+fn generate_hardware_software_mix_error(product_id: &str) -> ValidationError {
+    ValidationError {
+        message: format!(
+            "Product '{product_id}' contains serial_numbers or model_numbers but lacks a valid product path. This indicates a potential hardware and software mix in the product tree."
+        ),
+        instance_path: format!("/product_tree/.../product/id='{product_id}'"),
```
peinjoh: Minor: As the `instance_path` is used by consumers / users to highlight the errornous json section, we have so far used JSON pointer notation with the full path.

So this path should not contain '...' and I would argue it should end with `product_identification_helper/serial_numbers` / `.../model_numbers`. As there is currently no option to return an error with multiple paths, and to be as specific as possible with the paths, we could return up to two errors (if either serial_numbers and/or model_numbers) is present here instead.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+    use super::*;
+    use crate::csaf2_1::testcases::TESTS_2_1;
+
+    #[test]
+    fn test_test_6_2_31() {
+        let case_01_failing_example_1 = Err(vec![generate_hardware_software_mix_error("CSAFPID-908070601")]);
+
+        let case_11_valid_example_1 = Ok(());
+        let case_12_valid_example_2 = Ok(());
+        let case_13_valid_example_3 = Ok(());
```
peinjoh: Minor: We usually inline the Ok(())'s in the expect and provide a comment highlighting the content of the valid test cases (just a few words, to allow for discussions on test coverage). We are in the process of doing this for all of our own tests, but would appreciate that change to be made here :)

peinjoh (commented):
Hey there, thank you for your contribution! Here are some prelimary comments, I'll take a thourogh look later today and get back to you 👍

csaf-rs/src/validations/test_6_2_31.rs
tschmidtb51: > Minor: As the `instance_path` is used by consumers / users to highlight the errornous json section, we have so far used JSON pointer notation with the full path.

FYI: JSON Pointer is a recommendation per [CSAF 2.1](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#8-safety-security-and-data-protection-considerations-), which means that it is a requirement for this reference implementation.

> I would argue it should end with `product_identification_helper/serial_numbers` / `.../model_numbers`

Why? The test failure should be bound to the product as this is where it belongs to... Otherwise, someone could come to the conclusion to remove the serial / model number (which is obviously not intended).

peinjoh: I had misunderstood the goal of the test. After re-reading the spec, I now understand that the issue is not the presence of the production identification helper fields or the missing product path, but the implication of the combination of both. I agree that users should be pointed th the`.../product`, and that we should not nudge them to remove the product identifaction helper fields to get rid of the error. Thank you for bringing this up!

peinjoh:
Also: Please wire up the tests to the application by commenting in the validations.rs line 👍

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+            valid_path_references.push(sub_ref);
+        }
+    }
+
+    let mut errors: Vec<ValidationError> = vec![];
+
+    // 2. Use the built-in visitor pattern to safely check all full products across the entire tree
+    product_tree.visit_all_products_generic(&mut |product, _instance_path| {
+        if let Some(helper) = product.get_product_identification_helper() {
+            // now check if serial_nubers or model_numbers exist
```
peinjoh:
```suggestion
            // now check if serial_numbers or model_numbers exist
```

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Test 6.2.31: Hardware and Software
+pub fn test_6_2_31_hardware_and_software(doc: &impl CsafTrait) -> Result<(), Vec<ValidationError>> {
+    let Some(product_tree) = doc.get_product_tree() else {
+        return Ok(());
+    };
+
+    // 1. Gather all string references from the product paths cleanly
+    let mut valid_path_references: Vec<&str> = Vec::new();
+    let paths = product_tree.get_product_paths();
+
```
peinjoh, line 21: Two points here: As we only need to check for presence in the vec later, can this be deduplicated (with the included perf improvement) by using a HashSet?

I thought we have a utility function for this already. If not, extracting this as a utility function might be a good idea.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+        for sub_ref in path.get_subpath_product_references() {
+            valid_path_references.push(sub_ref);
+        }
+    }
+
+    let mut errors: Vec<ValidationError> = vec![];
+
+    // 2. Use the built-in visitor pattern to safely check all full products across the entire tree
+    product_tree.visit_all_products_generic(&mut |product, _instance_path| {
```
tziemek: You should call the non-generic version `visit_all_products` here to ensure all products are iterated depending on the csaf version. Currently both implementations call the `generic` function, but this may differ at some point.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+crate::test_validation::impl_validator!(csaf2_1, ValidatorForTest6_2_31, test_6_2_31_hardware_and_software);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::csaf2_1::testcases::TESTS_2_1;
+
+    #[test]
+    fn test_test_6_2_31() {
+        let case_01_failing_example_1 = Err(vec![generate_hardware_software_mix_error("CSAFPID-908070601")]);
```
peinjoh: I would argue that the test coverage can be improved here. There is currently no coverage of the `model_numbers` codepath, as the existing tests only concern `serial_numbers`. Similarly, there is no test products for serial_numbers and model_numbers being present at the same time (which is kind of trivial), and no test for what happens when there is no product_tree. We usually create a seperate issue for test coverage improvement, so this is not something you need to implement, but feel free to collect your thoughs on this on link them here 👍

peinjoh: Also: Test coverage could be extended to cover files that should produce multiple 6.2.31 warnings, i.e. with multiple offending products.

csaf-rs/src/validations/test_6_2_31.rs
tziemek: Actually we have a bug in the implementation which reports `relationships` for CSAF 2.1 documents #701 , which should be fixed first.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,77 @@
...
+    };
+
+    // 1. Gather all string references from the product paths cleanly
+    let mut valid_path_references: Vec<&str> = Vec::new();
+    let paths = product_tree.get_product_paths();
+
+    for path in paths {
+        let beg_ref = path.get_beginning_product_reference();
+        valid_path_references.push(beg_ref);
+
```
tziemek: A `full_product_name` in a `product_path` with the required `product_identification_helpers` would pass as valid here, despite being not referenced in another `product_path`. Please clarify if this is as intended, otherwise remove these from the valid paths so they are reported.
Also, please add a supplementary test to cover this.

tziemek (changes requested):
Please also update the README and validation.rs file(-s).

csaf-rs/src/validations/test_6_2_31.rs
tziemek: We have `get_relationships_product_references` already, but here the product_full_name are also included, which is, i think, not intended (see my comment).

georgbramm:
thanks for your valuable feedback, i will try to incorporate the requested changes by next week, as i am currently on a business trip

georgbramm:
i pushed a fix that hopefully resolves your requested changes

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -1,49 +1,53 @@
...
+    }
 
-        for sub_ref in path.get_subpath_product_references() {
-            valid_path_references.push(sub_ref);
-        }
+    // Track every product id explicitly defined inline inside relationships
+    for rel in product_tree.get_product_paths() {
+        let inline_product = rel.get_full_product_name();
+        valid_path_references.insert(inline_product.get_product_id().to_string());
     }
```
tziemek: This again adds the products, not the references, from the `product_paths` into the hashmap which will also be discovered be `visit_all_products`, because a product defined in `product_paths` is also a product. This would result in finding itself in the loop further down and thus ignoring it. 
If you think otherwise, please add a testcase which shows this scenario. If there is doubt about these products being relevant here, please open an issue with the OASIS TC.

georgbramm: You're right. A product shouldn't implicitly validate itself just by being defined inline as a full product node; it strictly needs to be target-referenced elsewhere within the path relationships or product groups to be considered structurally anchored. I have removed the secondary loop so that inline products are evaluated correctly against the reference map.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,75 @@
...
+    let Some(product_tree) = doc.get_product_tree() else {
+        return Ok(());
+    };
+
+    let mut errors: Vec<ValidationError> = vec![];
+
+    // 1. Gather all legitimate string references from groups and relationship paths
+    let mut valid_path_references: HashSet<String> = HashSet::new();
+
+    for (id, _) in product_tree.get_all_product_references() {
```
tziemek: Sorry to be nitpicky, but I think this is unclear.
We have to cover all 3 paths defined in the spec, which is covered by `visit_all_products` which returns all relevant products. -> Filter based on `serial_number` and `model_number`.
On the reference side, we should only use the references within the `product_paths`, so there has to be one product_path defined where each product is either the beginning or part of the subpaths.
So we shouldn't use `get_all_product_references` because this would include those from the product_groups. `get_relationships_product_references` should be the right one.

README.md
```diff
@@ -325,4 +325,4 @@ This creates TypeScript + WASM output in `wasm/`.
 
 | Test specification | 2.0                | 2.1 (experimental) |
 | --- |--------------------|--------------------|
-| 6.3.1 |  |  |
+| 6.3.1 | ⭕ | ✅ |
```
tziemek: This looks wrong

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: Not nitpicky at all—that's a critical distinction. Test 6.2.31 strictly checks for missing product paths (relationships), so allowing product group references to validate a hardware node introduces a false negative. I've switched the collector to use get_relationships_product_references() instead.

README.md
georgbramm: could you explain why? i though ⭕ means "Not applicable" ? as stated on line 230 in the README.md?

georgbramm: ahhh, now i see =) sorry... number was incorrect ^^

csaf-rs/src/validations/test_6_2_31.rs
tziemek: To resolve this conversation, a reference within a product group should not be considered here, only within a product path. But the current code now reflects already.

tziemek:
Please add some supplementary test cases for everything that was unclear/discussed in this PR.

georgbramm:
Last commit introduced a dedicated suite of supplementary test fixtures to resolve all edge-case validation behavior requested during the code review for Rule 6.2.31 (Hardware/Software Mix). These test scenarios are explicitly organized within a dedicated, manual test function (test_test_6_2_31_edge_cases). Isolating these structural, complex data matrix setups avoids interference with the rigid schema formatting assumptions used by the pre-generated test runner framework.

Rule 6.2.31 Supplementary Test Cases
Missing Product Tree Context (Case 01): Verifies that advisories completely omitting the optional product_tree structure return an Ok status rather than triggering runtime panics or structural validation failures.

Self-Validation Prevention (Case 02): Validates that standalone product entries containing conflicting hardware/software identifiers are caught, ensuring a product cannot implicitly clear its own requirement checks without valid external reference paths.

Model Numbers Code Path (Case 03): Adds targeted execution coverage specifically evaluating the model_numbers validation branch, ensuring full parity with existing serial_numbers tests.

Simultaneous Identifier Arrays (Case 04): Assures that the simultaneous presence of both serial_numbers and model_numbers inside a single configuration block is parsed and evaluated safely.

Multiple Offending Products (Case 05): Verifies full iteration and error accumulation mechanics by demonstrating that the validation engine captures all violation instances across multiple failing products instead of short-circuiting at the first failure.

tziemek:
> Last commit introduced a dedicated suite of supplementary test fixtures to resolve all edge-case validation behavior requested during the code review for Rule 6.2.31 (Hardware/Software Mix). These test scenarios are explicitly organized within a dedicated, manual test function (test_test_6_2_31_edge_cases). Isolating these structural, complex data matrix setups avoids interference with the rigid schema formatting assumptions used by the pre-generated test runner framework.
> 
> Rule 6.2.31 Supplementary Test Cases Missing Product Tree Context (Case 01): Verifies that advisories completely omitting the optional product_tree structure return an Ok status rather than triggering runtime panics or structural validation failures.
> 
> Self-Validation Prevention (Case 02): Validates that standalone product entries containing conflicting hardware/software identifiers are caught, ensuring a product cannot implicitly clear its own requirement checks without valid external reference paths.
> 
> Model Numbers Code Path (Case 03): Adds targeted execution coverage specifically evaluating the model_numbers validation branch, ensuring full parity with existing serial_numbers tests.
> 
> Simultaneous Identifier Arrays (Case 04): Assures that the simultaneous presence of both serial_numbers and model_numbers inside a single configuration block is parsed and evaluated safely.
> 
> Multiple Offending Products (Case 05): Verifies full iteration and error accumulation mechanics by demonstrating that the validation engine captures all violation instances across multiple failing products instead of short-circuiting at the first failure.

Case 01: looks fine, it's the obvious case
Case 02: generally looking fine, but the description says it prevents self validation. This is only possible within the product paths, not through full_product_names. Only the description is wrong, the case is still needed.
Case 03: looks fine, but then basically then same as case 02
Case 04: looks fine
Case 05: looks fine, the reasoning is confusing me. If you would add testfiles, this behavior is implicit as you would expect concrete results, not only an error.

Missing: case for a product declared within the product_path which has a serial or model number.

As mentioned in the other PR. Please add supplementary test files which can then be passed to the oasis tc.
If you need help with that, contact me.

georgbramm:
I have updated the pull request to resolve the CSAF 6.2.31 validation issues, including the necessary JSON schema corrections and the requested supplementary test coverage. The updated test suite now includes five specific edge cases to ensure the validator correctly identifies products lacking a valid product path:

Case 1: Tests a standard nested product that fails due to a lack of branch category distinction.
Case 2: Tests a 'Model Only' product to verify failure when the branch lacks hardware/software context.
Case 3: Tests 'Simultaneous' identifiers (serial and model) to ensure ambiguity is flagged in versioned branches.
Case 4: Tests an 'Isolated Self-Val' component to confirm that even self-contained products require valid hierarchy anchoring.
Case 5: Acts as a catch-all to ensure consistent rule application for unanchored identifiers.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
```diff
@@ -0,0 +1,77 @@
...
+  "$schema": "https://docs.oasis-open.org/csaf/csaf/v2.1/schema/csaf.json",
+  "document": {
+    "category": "csaf_security_advisory",
+    "csaf_version": "2.1",
+    "distribution": { "tlp": { "label": "CLEAR" } },
+    "publisher": {
+      "category": "vendor",
+      "name": "Test Publisher",
+      "namespace": "https://csaf.io"
+    },
```
tziemek, line 15: Please use the same publisher info as in the existing supplementary test files.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
```diff
@@ -0,0 +1,77 @@
...
+    "title": "OASIS Hardware Software Mix Baseline plus Edge Cases",
+    "tracking": {
+      "current_release_date": "2026-07-07T12:00:00Z",
+      "id": "CSAF-TEST-6-2-31-OASIS-01",
```
tziemek: `id` should be consistent with the file name.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
```diff
@@ -0,0 +1,77 @@
...
+        "branches": [
+          {
+            "category": "product_version",
+            "name": "1.0",
+            "product": {
+              "product_id": "CSAFPID-908070601",
+              "name": "Product 1",
+              "product_identification_helper": { "serial_numbers": ["SN-1"] }
+            }
+          },
```
tziemek: `serial_numbers` is already covered by the official test case 01.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
```diff
@@ -0,0 +1,77 @@
...
+          {
+            "category": "product_version",
+            "name": "5.0",
+            "product": {
+              "product_id": "CSAFPID-FIFTH-CASE",
+              "name": "Product 5",
+              "product_identification_helper": { "serial_numbers": ["SN-5"], "model_numbers": ["MOD-5"] }
+            }
+          }
```
tziemek: What is the difference between Product 3, 4 and 5?

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
tziemek, line 1: Please also format the JSON file.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -36,14 +36,30 @@ pub fn test_6_2_31_hardware_software_mix(doc: &impl CsafTrait) -> Result<(), Vec
...
+
+            if has_serial || has_model {
+                if !valid_path_references.contains(product_id) {
+                    let instance_path = format!("/product_tree/full_product_names/{idx}");
+                    errors.push(generate_hardware_software_mix_error(product_id, &instance_path));
+                }
+            }
+        }
+    }
+
```
tziemek: `product_tree.get_full_product_names()` is already covered by `product_tree.visit_all_products(...)`. Why do we need this additional iteration?

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
+            // 2. Model Only: Fails validation as branch lacks hardware/software context
+            generate_hardware_software_mix_error("CSAFPID-MODEL-ONLY", "/product_tree/branches/0/branches/1/product"),
+            // 3. Simultaneous: Fails as concurrent IDs create ambiguity in versioned branches
+            generate_hardware_software_mix_error("CSAFPID-SIMULTANEOUS", "/product_tree/branches/0/branches/2/product"),
+            // 4. Isolated Self-Val: Fails as branch nesting fails to qualify as a valid leaf
+            generate_hardware_software_mix_error("CSAFPID-SELF-VAL", "/product_tree/branches/0/branches/3/product"),
+            // 5. Fifth Case: Fails due to unanchored product identification in hierarchy
+            generate_hardware_software_mix_error("CSAFPID-FIFTH-CASE", "/product_tree/branches/0/branches/4/product"),
+        ];
+        s01.sort_by_key(|e| e.instance_path.clone());
```
tziemek: Again, sorting should not be necessary.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
-            Ok(()),
-            // Case 12: Valid - Product has no identification helper field configuration
-            Ok(()),
-            // Case 13: Valid - Pure software components without model or serial signatures
-            Ok(()),
-        );
-    }
+        // S01: Edge cases
+        let mut s01 = vec![
+            // 1. Nested Product: Fails because it lacks a clear branch category distinction
```
tziemek: What is meant by that? That there is `product_name` branch category missing? This doesn't concern this test.
Also it would be true for the other four cases listed here.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
+            generate_hardware_software_mix_error("CSAFPID-908070601", "/product_tree/branches/0/branches/0/product"),
+            // 2. Model Only: Fails validation as branch lacks hardware/software context
```
tziemek: The branch is not lacking the hardware/software context, a product path is missing.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
+            generate_hardware_software_mix_error("CSAFPID-MODEL-ONLY", "/product_tree/branches/0/branches/1/product"),
+            // 3. Simultaneous: Fails as concurrent IDs create ambiguity in versioned branches
```
tziemek: This description seems wrong. There is model and serialnumber present, which by itself is fine, but the product path is missing. There is no ambiguity.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
+            generate_hardware_software_mix_error("CSAFPID-SIMULTANEOUS", "/product_tree/branches/0/branches/2/product"),
+            // 4. Isolated Self-Val: Fails as branch nesting fails to qualify as a valid leaf
```
tziemek: This makes no sense, as it's the same data as case 03 and thus adds no additional value.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -56,175 +72,36 @@ mod tests {
...
+            generate_hardware_software_mix_error("CSAFPID-SELF-VAL", "/product_tree/branches/0/branches/3/product"),
+            // 5. Fifth Case: Fails due to unanchored product identification in hierarchy
```
tziemek: This makes no sense, as it's the same data as case 03 and thus adds no additional value.

tziemek (changes requested):

georgbramm:
Fixed: Synchronize publisher info and tracking ID, remove redundant test cases, and refine error messaging for missing product paths

tziemek:
> Fixed: Synchronize publisher info and tracking ID, remove redundant test cases, and refine error messaging for missing product paths

Please answer and/or address the open discussions in this PR.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: I’ve updated the JSON to treat the identification helper purely as a trigger for the rule. The focus of this test case is specifically the missing product path in the branch hierarchy, rather than the serial number itself, which I've retained only to ensure the validator evaluates the product's identification requirement.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: Agreed. I have removed these products to avoid redundancy. The test now focuses on a single, clear example of a missing product path in a branch, which is the core concern of 6.2.31.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You're correct, my terminology was imprecise. I've updated the comment to reflect that the failure is caused by a missing product path in the branch hierarchy.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: I have update it to:
    "publisher": {
      "category": "other",
      "name": "CSAF-RS Test Files",
      "namespace": "https://github.com/csaf-rs/csaf/tree/main/type-generator/assets/tests"
    },

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: i have update it to the filename

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You're absolutely right. I mistakenly assumed visit_all_products only targeted branch nesting. I've removed the redundant iteration over full_product_names and simplified the logic to rely entirely on the visitor, which correctly traverses the entire tree. Thanks for the catch!

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You are right. I removed sorting.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: Thanks for the correction. I have updated the comment to explicitly state that the product path is missing.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You are right, there is no ambiguity here. I reduced the test cases to one anyhow.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: Point taken. These were indeed redundant. I have simplified the test suite to a single case that demonstrates the missing product path requirement.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: Point taken. These were indeed redundant. I have simplified the test suite to a single case that demonstrates the missing product path requirement.

georgbramm:
I hope I answered and addressed all open problems in this pr correctly.

georgbramm:
i tried to address all comments, i fixed json format and i fixed json key sorting (and i fixed the conflict)

georgbramm:
fixed: cargo clippy warnings

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
tziemek, line 1: This is currently the same case as the one provided by the TC (Serial Number within branches).
I would expect too see cases for the different mentioned sources
```
  /product_tree/branches[](/branches[])*/product -> for serial_number already covered by TC 
  /product_tree/full_product_names[]
  /product_tree/product_paths[]/full_product_name
```
and for the model_number. 
Maybe there are also some valid cases missing.

tziemek (changes requested):

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: 554bbae extends the validation of CSAF test 6.2.31 to ensure comprehensive coverage across all product tree locations, including /product_tree/branches[], /product_tree/full_product_names[], and /product_tree/product_paths[]/full_product_name. The logic has been unified to strictly enforce consistency for both serial_number and model_number attributes across these sources.
I needed to extend produc_tree_trait.rs by get_product_path_prefix(&self) to get the version specific path.

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    let mut valid_path_references = HashSet::new();
+
+    for path in product_tree.get_product_paths() {
+        valid_path_references.insert(path.get_beginning_product_reference().to_owned());
+
+        for next in path.get_subpath_product_references() {
+            valid_path_references.insert(next.to_owned());
+        }
+    }
```
tziemek: Why revert back to doing this manually where there already was `get_relationships_product_references` used?

csaf-rs/src/validations/test_6_2_31.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Check Product Paths anchors
+    let prefix = product_tree.get_product_path_prefix();
+    for (i, rel) in product_tree.get_product_paths().iter().enumerate() {
+        check_product(
+            rel.get_full_product_name(),
+            &format!("{prefix}/{i}/full_product_name"),
+            &valid_path_references,
+            &mut errors,
+        );
+    }
```
tziemek: Why revert back to manually doing this instead of `visit_all_products`?

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
```diff
@@ -0,0 +1,120 @@
...
+        "product": {
+          "name": "B1",
+          "product_id": "CSAFPID-B1",
+          "product_identification_helper": {
+            "serial_numbers": [
+              "SN"
+            ]
+          }
+        }
+      },
```
tziemek: Product_tree -> branch -> product -> serial_numbers  is covered by the official 01 test case and can be left out here.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
tziemek: Nice addition to cover all relevant paths for invalid cases.

Maybe add another valid case (additional file) for a software (with a file_hash) not being used in a product_path, as this is not covered by official test files.
Also maybe add a valid case for a product with a serial number defined within the product_paths, which is again used in another product_path (this would likely not be the intended use but may come from a migration from csaf 2.0).

tziemek (changes requested):

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
oxisto: In the interest of looking at the overall effort spent on this PR, I would suggest to do this separately. This does not seem to be specific for this test case

tziemek: I'm fine with that, but then please add an issue for these cases that we don't forget to add them later.
Still, there are other comments open.

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You are right. thats better. 60e647d addresses this

csaf-rs/src/validations/test_6_2_31.rs
georgbramm: You are right. thats better. 60e647d addresses this

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-31-s01.json
georgbramm: I oversaw this =( I removed it from the test case in 21abbbb

georgbramm:
after the merge of main i had to adopt test_6_1_02.rs 
as paths differ between v2.0 (relationships) and v2.1 (product_paths)

tziemek (approved):

PR #698, opened 2026-06-23
feat: implement validation test 6.2.32 use of Same Product Identifica…

### LINKED ISSUE #318: Implement 6.2.32 Use of Same Product Identification Helper for Different Products for CSAF 2.1

georgbramm:
To fulfill the CSAF 2.0/2.1 specification for Test 6.2.32 (Use of Same Product Identification Helper for Different Products), I implemented a complete validation rule in src/validations/test_6_2_32.rs that ensures identifying product metadata is pairwise disjoint across all distinct products.

What Was Done

- Comprehensive Path Traversal: Integrated the framework's native visit_all_products_generic visitor engine. This automatically sweeps all valid paths defined in the spec, including recursive nested branches[], flat full_product_names[], and product_paths[] structures without requiring manual pointer-chasing.

- Pairwise Disjoint Set Validation: Implemented tracking maps (HashMap<String, String>) for every distinct helper property category: purls (Package URLs), skus (Stock Keeping Units), serial_numbers, model_numbers, hashes (Cryptographic file hashes)

- Object Deep-Equality with Order Independence: Built a robust serialization strategy for complex array properties whose items are JSON objects (such as hashes). The implementation extracts the filename and maps nested hashes into a deterministic format (Algorithm:HashValue), then explicitly sorts the array signatures before structural comparison. This guarantees that deep-equality checks pass successfully even if keys or items appear out of order.

- Error Reporting Alignment: Configured the rule to emit a detailed ValidationError whenever an identical identifier is found bound to multiple distinct product_id values, pinpointing the precise helper property path layout.

- Test Suite Stabilization: Aligned the unit tests with the automated test fixtures (case_01, case_02, case_03), verifying that both intentional failures (like the nested branch serial/model number collisions in Examples 1 & 2) and valid layouts pass cleanly against the standard framework mock runner.


csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,191 @@
...
+    category: &str,
+    value: &str,
+    product_id_1: &str,
+    product_id_2: &str,
+) -> ValidationError {
+    ValidationError {
+        message: format!(
+            "The Product Identification Helper property '{category}' contains a duplicate value '{value}' used across multiple distinct products ('{product_id_1}' and '{product_id_2}'). Properties must be pairwise disjoint."
+        ),
+        instance_path: format!("/product_tree/.../product_identification_helper/{category}"),
```
peinjoh: As in #696, this should be in JSON pointer notation with full path.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,191 @@
...
+    let mut seen_hashes: HashMap<String, String> = HashMap::new();
+
+    product_tree.visit_all_products_generic(&mut |product, _instance_path| {
+        let product_id = product.get_product_id().to_string();
+
+        if let Some(helper) = product.get_product_identification_helper() {
+            // Check PURLs
+            if let Some(purls) = helper.get_purls() {
+                for purl in purls {
+                    let purl_str = format!("{purl:?}");
```
peinjoh: This will use the debug output of `CsafPurl` enum returned by get_purls. This can be a) `Valid` or `Invalid`, depending on if the string could be parsed as a purl. The Valid struct provides a `normalized_purl` field, which might be helpful for the purpose of this test. 

I think it would be interesting to discuss here if this test should consider purl normalization and how invalid purls (that break 6.1.13) should be handled, or if a simple comparison of the original strings is sufficient.

peinjoh (commented):
Two more general comments: I think (and please correct me if I'm overlooking something) that some of the to_string() allocs are not necessary here. Is there a reason we are for example re-allocating the serial number as sn_str before using it as a hashmap key, and can this also be done without the to_string() call?

Also, the test coverage only seems to concern serial_numbers and model_numbers right now. I would argue this should be extended to als PIH's, if you have some thoughs here from your work on this task, feel free to open a ticket :)

csaf-rs/src/validations/test_6_2_32.rs
tziemek: We should probably reference the `product` here and rephrase the message.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,191 @@
...
+        // Case 02 contains a duplicate model number
+        let case_02_failing_example_2 = Err(vec![generate_duplicate_helper_error(
+            "model_numbers",
+            "143-D-354",
+            "CSAFPID-908070601",
+            "CSAFPID-908070602",
+        )]);
+
+        // Case 03 is an invalid example that yields 2 model number validation errors
+        let case_03_failing_example_3 = Err(vec![
```
tziemek: This case should report 3 messages, one for each product which has a duplicate. Probably don't reference the "other" product in the message, as we can't know which one is the correct one.

tziemek (changes requested):

georgbramm:
thanks again for your valuable feedback, i will try to incorporate the requested changes by next week, as i am currently on a business trip

georgbramm:
Hopefully fixes Issue #318

georgbramm:
Case 03 Messaging & Cardinality::
If three distinct products share the exact same identifier, all three are violating the pairwise disjoint rule. Instead of tracking the "first" product ID and naming it in the error message, I grouped by the metadata value, counted the occurrences, and if a value appears more than once, emited a clean error for every single product containing it without making assumptions.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,195 @@
...
+            for hash_obj in helper.get_hashes() {
+                let filename = hash_obj.get_filename();
+                let mut inner_signatures: Vec<String> = hash_obj
+                    .get_file_hashes()
+                    .iter()
+                    .map(|fh| format!("{:?}:{}", fh.get_algorithm(), fh.get_hash()))
+                    .collect();
+                inner_signatures.sort();
+
+                let hash_key = format!("file:{filename};hashes:{}", inner_signatures.join(","));
```
tziemek: If there is one product with a SHA256 file hash, and another product with the same filename and a SHA256 and SHA512 hash, they won't match but I think they should as it's just a difference in which file hashes are used. If only one matches for the same filename, we should fail here. Please add a supplementary test case for this, also for the other product identification helpers.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,195 @@
...
+                "/product_tree/branches/0/branches/0/branches/0/product",
+            ),
+            generate_duplicate_helper_error(
+                "serial_numbers",
+                "143-D-354",
+                "CSAFPID-908070602",
+                "/product_tree/branches/0/branches/0/branches/1/product",
+            ),
+        ];
+        case_01_errors.sort_by_key(|e| e.instance_path.clone());
```
tziemek: Why is sorting necessary here?

tziemek (changes requested):

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: Since we collect product occurrences using a HashMap, the internal iteration order of the errors is non-deterministic across test runs. Because the .expect() test harness macro validates strict index-by-index positional array equality, sorting the expected vectors in the test blocks by instance_path guarantees deterministic passes without testing bugs.

georgbramm:
i pushed a fix that hopefully resolves your requested changes

csaf-rs/src/validations/test_6_2_32.rs
tziemek: Sorry this is not correct. For the `expect` macro the cases itself must be in the correct order, but the expected errors within each don't have to be, see https://github.com/csaf-rs/csaf/blob/b723aecec76c6155b0b81d15ee13c17676c5bf7e/csaf-rs/src/test_result_comparison.rs#L37

csaf-rs/src/validations/test_6_2_32.rs
tziemek: @georgbramm 
In the `ValidPurl` case, this would look like
```
Valid(ValidPurl { original_purl: "pkg:generic/Example-Company-Product-A@1.0?type=product", normalized_purl: "pkg:generic/Example-Company-Product-A@1.0?type=product", base_without_qualifiers: "pkg:generic/Example-Company-Product-A@1.0" })
```
which is not fitting for a hashmap key. You could use the `normalized_purl` in the valid case, and `original_purl` in the invalid case, or `original_purl` in both cases. Either way, we shouldn't use the `Debug` output for hash keys.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,197 @@
...
+            }
+
+            // Collect Hashes
+            for hash_obj in helper.get_hashes() {
+                let filename = hash_obj.get_filename();
+
+                // Instead of one combined string, track each inner hash independently:
+                for fh in hash_obj.get_file_hashes() {
+                    let specific_hash_key =
+                        format!("file:{filename};alg:{:?};value:{}", fh.get_algorithm(), fh.get_hash());
```
tziemek: We shouldn't use the `:?` debug variant here too. Either use specific data or implement `Display`.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,197 @@
...
+                        .entry(specific_hash_key)
+                        .or_default()
+                        .push((product_id.clone(), path_str.clone()));
+                }
+            }
+        }
+    });
+
+    // 2. Process groups and emit independent errors for every product variant that collides
+    let mut process_violations = |groups: HashMap<String, Vec<(String, String)>>, category: &str| {
```
tziemek: This function should probably be a "real" private function. This would reduce the implicit access to `errors` and make the function itself testable.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,197 @@
...
+    process_violations(purl_groups, "purls");
+    process_violations(sku_groups, "skus");
+    process_violations(sn_groups, "serial_numbers");
+    process_violations(mn_groups, "model_numbers");
+    process_violations(hash_groups, "hashes");
+
+    // Sort and deduplicate to ensure deterministic error returns
+    // even if multiple hashes for the same file collide simultaneously.
+    errors.sort_by(|a, b| a.instance_path.cmp(&b.instance_path).then(a.message.cmp(&b.message)));
+    errors.dedup_by(|a, b| a.instance_path == b.instance_path && a.message == b.message);
```
tziemek: `ValidationError` is a struct with `PartialEq` so comparing the fields on its own should not be necessary. Maybe a simple `dedup` or other way to ensure uniqueness is simpler.
Also the `sort_by` should not be necessary, as the test comparison ignores ordering.

csaf-rs/src/validations/test_6_2_32.rs
tziemek: Additionally, for algorithm and hash value, we should use lowercase here. The schema will prohibit uppercase letters in these parts, but this test shouldn't rely on the schema test to run.

georgbramm:
Excellent callouts.

1) Switched out the {:?} debug string generation and replaced it with clean Display matching, forcing everything to .to_lowercase() to insulate the hashing tests from casing deviations regardless of schema enforcement rules.

2) Refactored the process_violations closure out into an independent, pure private helper function to make it fully testable.

3) Cleaned up the error deduplication by using simple .dedup() backed by ValidationError's native PartialEq implementation, and completely removed the redundant sorting step since the test suite runner handles random order matching natively.

tziemek:
Please also add some supplementary test cases for every edge case discussed here.

georgbramm:
I added three test cases for CSAF validation rule 6.2.32. The added tests verify that product identification helper metadata remains pairwise disjoint across all distinct entries within the product tree structure. These edge cases were implemented in a dedicated, isolated test function (test_test_6_2_32_edge_cases) to cleanly isolate manually structured validation payloads from the automatic schema formatting expectations of the pre-generated test matrix runner. 

The test suite evaluates three specific conditions:

Case 01: Duplicate Cryptographic Hash Detection (Failure)
Verifies that the validation engine flags an error when an identical cryptographic hash string (sha256 value) is declared across two separate product definitions (CSAFPID-HASH-FAIL-1 and CSAFPID-HASH-FAIL-2).

Case 02: Duplicate Package URL (PURL) Detection (Failure)
Verifies that the validation engine flags an error when an identical Package URL (pkg:npm/csaf-validator@0.5.1) is declared across two separate product definitions (CSAFPID-PURL-FAIL-1 and CSAFPID-PURL-FAIL-2).

Case 03: Pairwise Disjoint Identifiers (Success)
Verifies that the validation engine returns an Ok status when separate products declare entirely unique, non-overlapping hashes and PURLs, ensuring that valid configurations pass the rule without raising false positives.

tziemek:
> I added three test cases for CSAF validation rule 6.2.32. The added tests verify that product identification helper metadata remains pairwise disjoint across all distinct entries within the product tree structure. These edge cases were implemented in a dedicated, isolated test function (test_test_6_2_32_edge_cases) to cleanly isolate manually structured validation payloads from the automatic schema formatting expectations of the pre-generated test matrix runner.
> 
> The test suite evaluates three specific conditions:
> 
> Case 01: Duplicate Cryptographic Hash Detection (Failure) Verifies that the validation engine flags an error when an identical cryptographic hash string (sha256 value) is declared across two separate product definitions (CSAFPID-HASH-FAIL-1 and CSAFPID-HASH-FAIL-2).
> 
> Case 02: Duplicate Package URL (PURL) Detection (Failure) Verifies that the validation engine flags an error when an identical Package URL (pkg:npm/csaf-validator@0.5.1) is declared across two separate product definitions (CSAFPID-PURL-FAIL-1 and CSAFPID-PURL-FAIL-2).
> 
> Case 03: Pairwise Disjoint Identifiers (Success) Verifies that the validation engine returns an Ok status when separate products declare entirely unique, non-overlapping hashes and PURLs, ensuring that valid configurations pass the rule without raising false positives.

Perhaps my request was not clear enough: we try to create csaf test files to improve the overall test coverage of the oasis csaf repository. These are called supplementary test cases and are located within the `type-generator` in this repo. I would have expected some testfiles to cover all relevant and/or edge cases.

Additionally, because I have only seen that now, your code doesn't cover all `product_identification_helper`s. Both code and testfiles should cover all cases. For the supplementary test files, you also can combine multiple errors into one file to keep it more simple.

georgbramm:
I have updated the test suite to address the comments and ensure full coverage of rule 6.2.32.

Key changes:

Comprehensive Test Asset: Added csaf-rs_csaf-csaf_2_1-6-2-32-s01.json to include intentional collisions for all ProductIdentificationHelper categories: hashes, purls, serial_numbers, model_numbers, and skus.

Refactored Test Code: Updated the test expectation vector to match the new comprehensive collision set, ensuring deterministic validation.

georgbramm:
please wait with the review... im gonna fix the json & tests according to 6.2.31 comments

georgbramm:
now its ready for a review.. hopefully =)

tziemek:
> now its ready for a review.. hopefully =)

Please address or answer the open discussions in this PR.

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: Thanks for the clarification, Tziemek! You're absolutely right—I misunderstood the comparison logic in test_result_comparison.rs. I have removed the manual sort_by and dedup_by calls, as the test runner handles the unordered comparison natively. This definitely keeps the test code much cleaner.

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: I adopted it to:
let alg_str = format!("{}", fh.get_algorithm()).to_lowercase();
is this ok? or should i add Display impl?

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: I removed all sort_by's and used dedup

georgbramm:
I hope i addressed all comments and requests.
I will now fix the conflics (on both prs)

georgbramm:
i tried to address all comments, i fixed json format and i fixed json key sorting (and i fixed the conflict)

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,228 @@
...
+    let mut errors: Vec<ValidationError> = vec![];
+
+    // Grouping tracking maps: map a unique token to a list of (product_id, base_instance_path)
+    let mut purl_groups: HashMap<String, Vec<(String, String)>> = HashMap::new();
+    let mut sku_groups: HashMap<String, Vec<(String, String)>> = HashMap::new();
+    let mut sn_groups: HashMap<String, Vec<(String, String)>> = HashMap::new();
+    let mut mn_groups: HashMap<String, Vec<(String, String)>> = HashMap::new();
+    let mut hash_groups: HashMap<String, Vec<(String, String)>> = HashMap::new();
+
+    // 1. Collect all occurrences using zero-copy / minimal lifetime allocations where possible
```
tziemek, line 56: product identification helpers in 2.1 are extended compared to 2.0:
```
 product_identification_helper:
      cpe: String.Pattern
      hashes: Sequence
      model_numbers: Sequence
      purls: Sequence
      sbom_urls: Sequence
      serial_numbers: Sequence
      skus: Sequence
      x_generic_uris: Sequence
```
We should cover all cases here. If necessary, please extend the trait model for the missing properties.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,228 @@
...
+    });
+
+    process_violations(purl_groups, "purls", &mut errors);
+    process_violations(sku_groups, "skus", &mut errors);
+    process_violations(sn_groups, "serial_numbers", &mut errors);
+    process_violations(mn_groups, "model_numbers", &mut errors);
+    process_violations(hash_groups, "hashes", &mut errors);
+
+    // Simplified deduplication leveraging implemented PartialEq trait
+    errors.dedup();
```
tziemek: A simple `dedup` only takes consecutive occurrences into account. Maybe a HashMap is an easier approach. An additional testcase for duplicated (or rather non-duplicate errors) would be useful.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-32-s01.json
tziemek, line 1: Like for 6.2.31, having coverage regarding `branches` and `product_paths` would be beneficial in my opinion.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-32-s01.json
```diff
@@ -0,0 +1,160 @@
...
+          "P5",
+          "P6",
+          "P7",
+          "P8",
+          "P9",
+          "P10"
+        ]
+      }
+    }
+  ]
```
tziemek: Vulnerabilities are irrelevant for this test. We should try to keep the file content at a minimum in order to have a clear picture.

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: fixed in 3412fea:
Key Technical Improvements:

Trait-Bound Type Access: Resolved "unknown field" compilation errors by introducing specialized trait bounds (GenericUriTrait, CpeTrait) on the ProductIdentificationHelperTrait associated types. This allows the validator to safely and statically access concrete data fields without modifying the core schema or using unsafe blocks.

Safe Static Dispatch: Replaced opaque trait object handling with static trait bounds, ensuring type-safe, performant field access across different CSAF versions.

Consolidated Validation Logic: Streamlined the grouping and duplicate-checking logic for all eight identification helper categories, ensuring accurate JSON path generation and consistent error reporting.

Test Suite Synchronization: Updated the 6.2.32 test cases to align with current schema-compliant path generation and key formatting, ensuring robust coverage for both simple and complex (nested) product identification structures.

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-32-s01.json
georgbramm: removed/fixed in a7fc66a

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-32-s01.json
georgbramm: also testing branches (adopted json and test code) now, in 49ecf08

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: i am not 100% sure if i addressed this correctly with b26c8ae
i added a new test case which verifies that duplicate helper values inside the *same* product are correctly ignored (expects pass).
Please correct me, if i got it wrong

csaf-rs/src/csaf/traits/vulnerabilities/product_ident_helper_trait.rs
```diff
@@ -41,10 +45,21 @@ pub trait ProductIdentificationHelperTrait {
 
     /// Returns the hashes associated with this product.
     fn get_hashes(&self) -> &Vec<Self::HashType>;
+
+    /// Returns the CPEs (Common Platform Enumeration).
+    fn get_cpes(&self) -> Option<Vec<Self::CpeType>>;
```
tziemek: This shouldn't be a `Vec` as there can only be one CPE within a single product identification helper.

csaf
tziemek, line 1: Please don't update the submodules within a feature PR.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,353 @@
...
+    // 1. Collect all occurrences using zero-copy / minimal lifetime allocations where possible
+    product_tree.visit_all_products(&mut |product, instance_path| {
+        let product_id = product.get_product_id().to_string();
+        let path_str = instance_path.to_string();
+
+        if let Some(helper) = product.get_product_identification_helper() {
+            // Collect PURLs
+            if let Some(purls) = helper.get_purls() {
+                for purl in purls {
+                    let key = format!("{purl:?}"); // format necessary due to enum debug representation
```
tziemek: Still there is no need for the debug string, we should use the original or even the normalized form here.

csaf-rs/src/validations/test_6_2_32.rs
```diff
@@ -0,0 +1,353 @@
...
+                        .entry(specific_hash_key)
+                        .or_default()
+                        .push((product_id.clone(), path_str.clone()));
+                }
+            }
+
+            // Collect CPEs
+            if let Some(cpes) = helper.get_cpes() {
+                for cpe in cpes {
+                    let key = format!("cpe:{}", cpe.as_str());
```
tziemek: Why are cpe, sbom, x_uri prefixed? As you have explicit groups for each category, there is no need for this prefix.

csaf-rs/src/csaf/traits/vulnerabilities/product_ident_helper_trait.rs
```diff
@@ -102,7 +139,19 @@ impl ProductIdentificationHelperTrait for HelperToIdentifyTheProduct21 {
...
     }
 
-    fn get_hashes(&self) -> &Vec<Self::HashType> {
-        self.hashes.as_ref()
+    fn get_hashes(&self) -> Option<Vec<Self::HashType>> {
+        self.hashes.clone()
+    }
+
+    fn get_cpes(&self) -> Option<Vec<Self::CpeType>> {
+        self.cpe.as_ref().map(|v| vec![v.clone()])
```
tziemek: self.cpe is a newtype around a string, so it shouldn't be a vector.

csaf-rs/src/csaf/traits/vulnerabilities/generic_uri_trait.rs
```diff
@@ -0,0 +1,18 @@
+use crate::csaf::traits::util::impl_str_field_getter;
+use crate::schema::csaf2_0::schema::GenericUri as GenericUri20;
+use crate::schema::csaf2_1::schema::GenericUri as GenericUri21;
+
+pub trait GenericUriTrait {
```
tziemek, line 5: Currently we have rather structs than traits, like with model_number etc., but I am not sure what way will be better in the future. Keep the trait for now.

tziemek (changes requested):

csaf
oxisto: It seems that the build failed with the asset check, so that probably would need to be removed if it's not extended for contributors to update the assets within the PR. See https://github.com/csaf-rs/csaf/actions/runs/29512181311/job/87668133727

tziemek: Why should we remove the asset check? I only stated to not update the assets within a PR which implements a feature. Doing a PR for only updating the assets, and thus potentially changing more than a single test, is totally fine.
The asset update is already open as #711  from renovate, so a rebase/merge would also handle that once it is merged and lead to less confusion.

oxisto: > Why should we remove the asset check? I only stated to not update the assets within a PR which implements a feature. Doing a PR for only updating the assets, and thus potentially changing more than a single test, is totally fine. The asset update is already open as #711 from renovate, so a rebase/merge would also handle that once it is merged and lead to less confusion.

Because it fails the CI pipeline, making the contributors think that something is wrong. Then they update the assets because the CI pipeline literally tells them to do it :)

It should at least not fail in a way that the regular build is not even executed.

milux: @tziemek I tried to help @georgbramm in sorting out an issue with the pipeline where the very same commit on this branch here caused `csaf-rs/src/csaf2_1/testcases.generated.rs` to change on my machine and in the CI pipeline when executing `cargo run --locked -p type-generator`, yet not on Georg's machine, even when he explicitly deleted the file to force regeneration. The command simply showed different behavior on our machines, likely because of some stale resources somewhere that haven't been updated properly for whatever reason.

This problem "went away" after asset update, with the undesired side effect that even more updates - as you correctly pointed out - became necessary to keep the pipeline from exploding.

So, disabling the asset update might be a starting point to prevent such issues, but there might be even more problems with stale generated files even after that which need fixing, yet certainly not within this branch...

Besides: Would it be technically possible to implement a check that prevents explicit submodule updates ahead of `main` within feature PRs to start with? :thinking:

tziemek: > > Why should we remove the asset check? I only stated to not update the assets within a PR which implements a feature. Doing a PR for only updating the assets, and thus potentially changing more than a single test, is totally fine. The asset update is already open as #711 from renovate, so a rebase/merge would also handle that once it is merged and lead to less confusion.
> 
> Because it fails the CI pipeline, making the contributors think that something is wrong. Then they update the assets because the CI pipeline literally tells them to do it :)
> 
> It should at least not fail in a way that the regular build is not even executed.

What are the consequences of updating the csaf submodule?
- there might me changed requirements in the standard and/or several tests
- there might be changed test files

Those have to be reflected in the codebase. This is done by 
- updating the assets (copying data from the submodule into this repository)
- generating the appropriate code (type-generator) 
- and most importantly changing the code according to changed requirements. 

We see this as an atomic operation, unless some changes are too big. In this case, the code should pass the buildpipeline and an issue should be created to change those specific areas. 
If we allow only updating the submodule reference and disabling the asset check, it would mean the next person to do an asset update (maybe just to make sure everything is up to date) would have to incorporate changes out of the blue. We don't want to pressure anyone into doing that, so the update process is an explicit task. Additionally, dependabot provides us with the updated submodules so we have a dedicated PR to look into these changes.
We'll be adding a section on how this mechanism works to our contributing guidelines / contributing documentation shortly, as we now see that this process is not as straightforward to first time contributors as it is to us.
We could make this a side job, so the build step runs, but the job fails anyways.

tziemek: > @tziemek I tried to help @georgbramm in sorting out an issue with the pipeline where the very same commit on this branch here caused `csaf-rs/src/csaf2_1/testcases.generated.rs` to change on my machine and in the CI pipeline when executing `cargo run --locked -p type-generator`, yet not on Georg's machine, even when he explicitly deleted the file to force regeneration. The command simply showed different behavior on our machines, likely because of some stale resources somewhere that haven't been updated properly for whatever reason.
> 
> This problem "went away" after asset update, with the undesired side effect that even more updates - as you correctly pointed out - became necessary to keep the pipeline from exploding.
> 
> So, disabling the asset update might be a starting point to prevent such issues, but there might be even more problems with stale generated files even after that which need fixing, yet certainly not within this branch...
> 
> Besides: Would it be technically possible to implement a check that prevents explicit submodule updates ahead of `main` within feature PRs to start with? 🤔

For having different generated files on different machines: either one of you may have (accidently) updated the submodule in their local copy (and run the asset update script), otherwise it is not possible to have different generated files. There are also no stale files in the repository, as the main build is working (which also checks for missing updates). If your mention of stale means, not up to date with the current main branch of the csaf repository, then you are right, but that's intentional (see above).

tziemek: > 
> Besides: Would it be technically possible to implement a check that prevents explicit submodule updates ahead of `main` within feature PRs to start with? 🤔

The asset check was our way to do that on technical level. But having it locally would be better.

milux: > For having different generated files on different machines: either one of you may have (accidently) updated the submodule in their local copy (and run the asset update script), otherwise it is not possible to have different generated files. There are also no stale files in the repository, as the main build is working (which also checks for missing updates). If your mention of stale means, not up to date with the current main branch of the csaf repository, then you are right, but that's intentional (see above).

I thought so, too. Therefore, we checked for that multiple times and found the submodules were on the same commit on both machines, yet the outcome was different still. For precisely that reason I assume a caching error, but since @georgbramm had no time to bother with deep root cause analysis, our evidence is now lost and we'll have to wait for the bug showing up again, if it does some day. :man_shrugging:

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: i removed the prefixes in 9b55364

csaf-rs/src/csaf/traits/vulnerabilities/product_ident_helper_trait.rs
georgbramm: i removed the Vec in 9b55364

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: i addressed this issue in 5f0c41a using match. I had to add a getter to PurlParseError for the Error case.

csaf-rs/src/validations/test_6_2_32.rs
georgbramm: using normalized now

tziemek (approved):

PR #699, opened 2026-06-23
Implementing 6.2.24 Usage of Non-Latest CWE Version

### LINKED ISSUE #310: Implement 6.2.24 Usage of Non-Latest CWE Version for CSAF 2.1

konradweiss:
Note: Some of the test cases that are expected to fail use a newer version than the current one. For now I have implemented them with the same Error message, which could be changed.

Resolves #310


konradweiss:
Closes #310

peinjoh:
Is this ready to be reviewed?

konradweiss:
> Is this ready to be reviewed?

Yes, and happy to get some advice for improvements :)

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::csaf2_1::testcases::TESTS_2_1;
+
+    #[test]
+    fn test_test_6_2_24() {
+        let case_01 = Err(vec![create_non_latest_cwe_error(
```
peinjoh:
```suggestion
        let case_01_cwe_version_before_latest = Err(vec![create_non_latest_cwe_error(
```

We try to go for speaking variable names for the failing cases. For these, it's kind of trivial to see whats wrong, but we found this helpful for more complex cases.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+            "4.13",
+            "/vulnerabilities/0/cwes/0",
+        )]);
+
+        let case_02 = Err(vec![create_non_latest_cwe_error(
+            "CWE-143",
+            "4.15",
+            "4.13",
+            "/vulnerabilities/0/cwes/0",
+        )]);
```
peinjoh: Same as above, I think a speaking var name would help. Also, I would like to point out the error message being generated here:

`Weakness 'CWE-143' uses non-latest CWE version 4.15 (latest: 4.13).`

This might be confusing to users, as the message does not indicate that the latest CWE version at the time of release is evaluated, and 4.15 is clearly later than 4.13. We could for example a) include a mention of the release date of latest or b) provide a different error message for when latest is "earlier" than the actual version of the CWE.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+            create_non_latest_cwe_error("CWE-287", "1.0", "4.13", "/vulnerabilities/0/cwes/2"),
+        ]);
+
+        let case_04 = Err(vec![
+            create_non_latest_cwe_error("CWE-158", "1.3", "4.13", "/vulnerabilities/0/cwes/0"),
+            create_non_latest_cwe_error("CWE-138", "2.1", "4.13", "/vulnerabilities/0/cwes/1"),
+            create_non_latest_cwe_error("CWE-318", "4.14", "4.13", "/vulnerabilities/1/cwes/0"),
+            create_non_latest_cwe_error("CWE-61", "4.15", "4.13", "/vulnerabilities/2/cwes/0"),
+        ]);
+
```
peinjoh, line 186: We try to provide a small comment on what the passing test cases do, to allow for a shared unterstanding and discussion of test coverage. 

For example:
```
// Case 11: 1 vuln, 1 correct cwe version, correction of case 01
// Case 12: 1 vuln, 1 correct cwe version, corrrection of case 02
// Case 13: 1 vuln, 3 correct cwe versions, correction of case 03
// Case 14: multiple vulns, multiple correct cwe versions, correction of case 04
```

Did you find any code paths / gaps during implementation? If so, feel free to open an issue so we can add a supplemental test for it :)

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+                let Some(latest) = latest_version else {
+                    continue;
+                };
+
+                if version != latest {
+                    errors.push(create_non_latest_cwe_error(
+                        &cwe_item.id,
+                        version,
+                        latest,
+                        format!("/vulnerabilities/{i_r}/cwes/{i_cwe}").as_str(),
```
peinjoh: I think we can safe a few string allocations by passing the indices and only having one format! call in the create_non_latest_cwe_error() function here :)

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+                // should have rejected the document; skip checking here.
+                let Some(version) = cwe_item.version.as_deref() else {
+                    continue;
+                };
+
+                // If we cannot determine the latest CWE version for the document,
+                // skip the check (no external CWE data available).
+                let Some(latest) = latest_version else {
+                    continue;
+                };
```
peinjoh: This check can live outside this loop I think, if the latest version is None, this test can not run. 

There are ongoing discussions how to handle this (test not being able to run, being skipped for various reasons, ...) upstream (one of them: https://github.com/oasis-tcs/csaf/issues/1231). We had this discussion as #409 in our repo.

Please add a comment on which under which coniditions latest_version can be None, which will help us choose the correct return type during that coming refactor.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+    // If the date is invalid, we cannot determine the latest CWE version — skip.
+    if !date.is_valid() {
+        return None;
+    }
+
+    // Convert to a date (UTC) and compare against the release dates stored in the CWE assets.
+    let doc_date: NaiveDate = match date {
+        CsafDateTime::Valid(v) => v.get_as_utc().date_naive(),
+        _ => return None,
+    };
```
peinjoh: These two checks are duplicate, as date.is_valid() is just `matches!(self, CsafDateTime::Valid(_))`, which you do in the second statement. 

Also, lets discuss if a) this check should be done in the main function, and this function should accept only `ValidCsafDatetime` structs, and also b if this function should be co-located with `CWE_ENTRIES`

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,127 @@
...
+    let current_release_date = tracking.get_current_release_date();
+
+    // Determine the latest CWE version available at document current_release_date.
+    let latest_version = get_latest_cwe_version_for_date(&current_release_date);
+
+    for (i_r, vulnerability) in vulnerabilities.iter().enumerate() {
+        if let Some(cwes) = vulnerability.get_cwe() {
+            for (i_cwe, cwe_item) in cwes.iter().enumerate() {
+                // Require the CWE item to include a version. If missing, the parser
+                // should have rejected the document; skip checking here.
```
peinjoh: This comment is slightly incorrect: CWE version is required for CSAF 2.1 and noneexistent for CSAF 2.0. So the CSAF 2.1 deserialization would have broken before this test could be executed, making this codepath practically unreachable.

This code path should be revisited if we implement #159 (although this test simply does not make sense for CSAF 2.0 and would probably be skipped altogether) and #571, where the strict requirements introduced by our generated type deserialization will not exist and this codepath could be visited.

peinjoh:
Lastly: A big thank you for your first contribution(s) to your project, if there are any further questions we are happy to help!

I've written a bit more lengthy comments to show you around some of the currently still rough edges of our codebase, my comments might also find their way into #700 or similar so others don't stumble over the same things.

Thanks again, and I look forward to the discussions and your potential future contributions 👍

konradweiss:
> Lastly: A big thank you for your first contribution(s) to your project, if there are any further questions we are happy to help!
> 
> I've written a bit more lengthy comments to show you around some of the currently still rough edges of our codebase, my comments might also find their way into #700 or similar so others don't stumble over the same things.
> 
> Thanks again, and I look forward to the discussions and your potential future contributions 👍

Thanks, for you review comments :) I will look to address them ASAP, which for me should be on Friday

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I added a comparison based on semantic versioning and reported the case that the used CWE is of a newer version with a new error message.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I documented these testcases as recommended above and added information on the versions that changed and therefore fulfilled the testcases.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: Can you take a look at my latest changes? We now have different error messages and I am not sure if we can still wrap them together.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: After revisiting it, I am now of the opinion that latest version information should not be missing if the initialization was done correctly and therefore I am now returning an error if the latest version is None.

I did also follow your suggestion and moved the checkout outside of the loop into an `if-let`

csaf-rs/src/validations/mod.rs
```diff
@@ -118,6 +118,7 @@ pub mod test_6_3_05;
 // pub mod test_6_3_06;
 // pub mod test_6_3_07;
 // pub mod test_6_3_08;
+pub mod test_6_2_24;
```
peinjoh: Minor: This file should be lexicgraphically sorted, please move this line appropriatly.

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Thanks, I'll give it a review shortly!

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: If the first check is a duplicate of the match, I would remove the `is_valid()` check. I need to use the match to get to the doc date. So if there is a way to circumvent that by getting a `ValidCsafDatetime` from get_current_release_date, that would solve the issue. I agree that if similar functionality is used elsewhere, we should extract it to another location, eg. helpers.rs.

What do you think? Should I leave it as is, by omitting `date.isValid()` or do you see the need to refactor as suggested in this PR?

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I for now updated the comment. The check itself is necessary for the extraction, but similar to the check that I moved out of the loop we could change this to return some error here. I, however, think that that is not necessary, because as you mentioned: we will never run into that error case in the future if I understood you correctly.

konradweiss:
Some parts of the build are failing right now, I will fix them and remark the PR as ready for review. I am sorry for the inconvenience.

csaf-rs/src/validations/mod.rs
konradweiss: Done

konradweiss:
> Some parts of the build are failing right now, I will fix them and remark the PR as ready for review. I am sorry for the inconvenience.

@peinjoh The issue was the renaming of `get_cwe` to `get_cwes` on main, so the build did not go through in the pipeline. Now it should all be working an ready for further review and comments :)

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+use crate::csaf_traits::{CsafTrait, DocumentTrait, TrackingTrait, VulnerabilityTrait};
+use crate::helpers::CWE_ENTRIES;
+use crate::validation::ValidationError;
+use semver::Version;
+
+fn create_non_latest_cwe_error(cwe: &str, version: &str, latest: &str, path: &str) -> ValidationError {
+    // Parsing both strings as semantic versions. CWE assets use two-part
+    // versions like "4.13". `semver::Version::parse` expects three parts in the version
+    // (major.minor.patch), we normalize the version by appending 0 until we have three segments
+    //  and preserver prerelease/build metadata.
```
peinjoh: Minor typo

```suggestion
    //  and preserve prerelease/build metadata.
```

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+    let parsed_latest = Version::parse(norm_latest.as_str());
+
+    let error_message = match (parsed_version, parsed_latest) {
+        (Ok(v), Ok(l)) => {
+            if v < l {
+                format!("Weakness '{cwe}' uses non-latest CWE version {version} (latest: {latest}).")
+            } else if v > l {
+                format!("Weakness '{cwe}' uses a future CWE version {version} (latest: {latest}).")
+            } else {
+                format!("Weakness '{cwe}' uses latest CWE version {version} but validation is incorrect.")
```
peinjoh: In which cases do you expect the else arm here to be executed? As far as I can see, in the test function, we initially check if `version != latest`, so this arm should be pratically unreachable, or am I missing something here? 

Same applies for the `_ => ` arm below I think.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+    let vulnerabilities = doc.get_vulnerabilities();
+    let mut errors: Vec<ValidationError> = Vec::new();
+
+    let tracking = doc.get_document().get_tracking();
+    let current_release_date = tracking.get_current_release_date();
+
+    // Determine the latest CWE version available at document current_release_date.
+    let latest_version = get_latest_cwe_version_for_date(&current_release_date);
+
+    // If we cannot determine the latest CWE version for the document,
```
peinjoh, line 115: This comment contradicts the actual implementation. Skipping would be returning an Ok(), but we are throwing an error.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+    let mut latest: Option<(&'static String, &NaiveDate)> = None;
+
+    for (version, (release_date, _)) in CWE_ENTRIES.iter() {
+        if *release_date <= doc_date && latest.as_ref().is_none_or(|l| *release_date > *l.1) {
+            latest = Some((version, release_date));
+        }
+    }
+
+    latest.map(|(version, _)| version)
+}
```
peinjoh: Site note: This function could be moved closer to the CWE_ENTRIES definition and made available as pub, I think this will be helpful elsewhere.

EDIT:  Using the function in the other tests is out of scope of this PR, please just open an issue for this.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: Yes, this should be unreachable until someone uses this function although the versions are matching. Because i have to return something it appeared logical to me to return this error, although it should not happen if the implementation uses the error generating function only when `version != latest`.

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+                    let Some(version) = cwe_item.version.as_deref() else {
+                        continue;
+                    };
+
+                    if version != latest {
+                        errors.push(create_non_latest_cwe_error(
+                            &cwe_item.id,
+                            version,
+                            latest,
+                            format!("/vulnerabilities/{i_r}/cwes/{i_cwe}").as_str(),
```
peinjoh: Minor: For consistency, please move the indices to the method header, and generate the instance path in the error gen function

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+            core = &s[..idx];
+            rest = &s[idx..];
+        }
+
+        let mut parts: Vec<&str> = core.split('.').collect();
+        while parts.len() < 3 {
+            parts.push("0");
+        }
+        format!("{}{}", parts.join("."), rest)
+    }
```
peinjoh: Please add some minimal unit test coverage to a) pin behaviour and b) provide example for other contributors

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+        },
+        // If parsing fails for either side, fall back to string comparison to avoid panics.
+        _ => {
+            if version < latest {
+                format!("Weakness '{cwe}' uses non-latest CWE version {version} (latest: {latest}).")
+            } else if version > latest {
+                format!("Weakness '{cwe}' uses a future CWE version {version} (latest: {latest}).")
+            } else {
+                format!("Weakness '{cwe}' uses latest CWE version {version} but validation is incorrect.")
```
peinjoh: We discussed a refactoring: Move the `if version != latest` check into this function, rename to `check_*`, with either result or optional return type. Then these else blocks can be `unreachable!`. And move the error generation into a seperate, minimal function, with enums for each of the 2 error format strings.

peinjoh (changes requested):

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Comparison has its own comment now for refactoring, but this comment is resolved. 👍

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: We discussed the refactoring and added a comment at the new codes location. This is resolved 👍

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I used an enum for the decision process and and extracted the error generation. This function now checks the version and removes duplicate string generation. The case where version is neither `version>latest` nor `version<latest` no error is returned.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I changed the comment as discussed

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I extracted the function to `helpers.rs` and created and issue to reuse it in the future.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I moved the indices to the version check, and there into the error generation function, similarly to #695

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: I added some small amount of test cases to show and pin the behavior. I did so, specifically for cwe version cases that are relevant in this test.

csaf-rs/src/validations/test_6_2_24.rs
konradweiss: As discussed i generated an Enum to reduce calls to format and used `Option<>` to avoid creating an unnecessary error.

konradweiss:
This branch and all its review comments should be addressed as discussed.

peinjoh:
Will review shortly 👍

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: I slightly disagree with you here on second thought. This test concerns specific cwes not having the latest version available at the current release date of the document.

It should not generate an error concerning our infrastructure being set up wrong. That being said, we don't have a validation ensuring that CWE_ENTRIES is filled at all right now, and until that is in place, this error being returned is fine. 

I'm creating an issue to add a compile-time check for this though, and if that is implemented, we can replace this else block with unreachable! 👍  (#753)

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Co-location looks good. I still favor the function accepting ValidCsafDatetime, specially as it is now more publically available. That refactoring can be done during #750 though, when we have a better understanding of how this function will be used across the lib. Fine for now  👍

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Fine as-is.

For further context and posterity, here is a bit more detail on how this test might act in the different future scenarios mentioned.

For CSAF 2.1 (which this test is currently only running for), this check is always passing, as cwe.version is mandatory via the schema. A document missing it would have failed deserialization.

We have discussed #159, which would allow CSAF 2.1 tests to run on CSAF 2.0 documents. Here, cwe has only 2 properties `id` and `name`, i.e. the cwe field has no concept of a version. The check would therefore always fail. We could try to infer the version by matching the id / name pairs against the cwe database, but is completly out of scope.

We are also activly working on #571. This would allow CSAF 2.0 and 2.1 tests to run on documents that would otherwise fail deserialization, i.e. partially broken / incomplete documents. Here, that check would actually "do something" as there could for example be a CSAF 2.1 document with mulitple cwes, some of them having a version, some of them without a version, and this test would still work.

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Looking good 👍

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,202 @@
...
+    let latest_version = get_latest_cwe_version_for_date(&current_release_date);
+
+    // If we cannot determine the latest CWE version for the document,
+    // we generate an error as the cwe list was not initialized correctly.
+    if let Some(latest) = latest_version {
+        for (i_r, vulnerability) in vulnerabilities.iter().enumerate() {
+            if let Some(cwes) = vulnerability.get_cwes() {
+                for (i_cwe, cwe_item) in cwes.iter().enumerate() {
+                    // Extract the CWE version that is mandatory for CSAF 2.1 but
+                    // optional for CSAF 2.0., this test is only relevant for the former
```
peinjoh:
```suggestion
                    // does not exist in CSAF 2.0, this test is only relevant for the former
```

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,202 @@
...
+    e_type: VersionMissmatch,
+    cwe: &str,
+    version: &str,
+    latest: &str,
+    i_r: usize,
+    i_cwe: usize,
+) -> ValidationError {
+    let error_message = match e_type {
+        VersionMissmatch::NonLatest => {
+            format!("Weakness '{cwe}' uses non-latest CWE version {version} (latest: {latest}).")
```
peinjoh:
```suggestion
            format!("Weakness '{cwe}' uses non-latest CWE version '{version}' (latest: '{latest}').")
```

csaf-rs/src/validations/test_6_2_24.rs
```diff
@@ -0,0 +1,202 @@
...
+        },
+        VersionMissmatch::Future => format!("Weakness '{cwe}' uses a future CWE version {version} (latest: {latest})."),
```
peinjoh:
```suggestion
        VersionMissmatch::Future => format!("Weakness '{cwe}' uses a future CWE version '{version}' (latest: '{latest}')."),
```

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Looks good functionality-wise, comment on possible refactoring improvement below.

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Enum and Option looks good. Naming-wise, the return type option does indicates that something optional is found / not found.

`check_version` does not clearly indicate what is found here, I would favor something more speaking like `check_for_non_latest_cwe_version`, then it would clearly indicate that a) Some(ValidationError) contains a ValidationError on a non-latest cwe version and b) None means no non-latest cwe version was found.

Otherwise, I would still argue the `if version != latest` check can be moved from the test function in here, with an early `return None`. Then the two else-blocks in `check_version` can be removed, as that case is handled in the function scope already, and we are sure that the versions are either `a<b` or `a>b`. 

Both are more stilistic issues and not blocking the merge, though.

peinjoh: Otherwise, PR looks fine beyond a few minor corrections / typos I added as suggestions 👍

konradweiss: I changed the function name to the proposed version and moved the inequality check to the function as an equality check that creates an early return

peinjoh (approved):
LGTM! Is there something left to do from your side here? Otherwise I'll merge 👍

csaf-rs/src/validations/test_6_2_24.rs
peinjoh: Perfect, thanks!

konradweiss:
> LGTM! Is there something left to do from your side here? Otherwise I'll merge 👍

No, this is finished :D

peinjoh:
Alright, thank your for your contribution! 👍

PR #712, opened 2026-07-18
test: hold the 6.1.13 PURL pipeline to the official purl-spec suite

gronke:
Every parse/roundtrip case of the pinned purl-spec submodule runs through the schema pattern and the packageurl parse; divergences live in per-version KNOWN_GAPS lists that fail on regressions and on stale entries alike. 2.0 misjudges 16 unique purls (4 spec-valid rejected, 12 type-invalid accepted); 2.1 misjudges 25 — its stricter schema pattern also rejects mixed-case types, which the purl specification accepts.


peinjoh:
Thanks, this is a great improvement to our test coverage. We'll be looking into the gaps, but this might take a bit 👍

peinjoh:
I approve of the changes in general, but I think there is a way to better fit this into our asset / build process.

So far, external submodules are included via our asset_update script, so the files relevant to our projects are copied (and therefore seperatly diffed) into the crates where they are needed, and we get pinged by dependabot when a diff exists.

So I would:
1) test and fix the inconsistencies you pointed out
2) extend our asset update mechanism to include the purl spec test suite json files
3) include the test suite json files, possibly with an "expected results" json to pin the "known gaps"
4) might move the purl tests into a seperate test crate entirely, with the option to add more authoritative sources in a similar fashion

gronke:
I've closed https://github.com/scm-rs/packageurl.rs/pull/31 again after noticing it was a duplicate and @ctron already decided against purl spec via git submodule.

Didn't want to publish/vendor yet another purl implementation, but built one that is fully compliant to the spec (and tested) for internal use. If you're interested, I'll put it on a GitHub repo separately. I personally think it's better for the ecosystem to let packageurl-rs decide on a testing strategy and then contribute to make it fully compliant and switch to strict enforcement via CI.

peinjoh:
I fully agree that the effort should be on packageurl-rs being improved, but I would also be interested in your implementation :)

As for CSAF, the standard is explicit in not doing "exact to spec" purl, as described [here](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#full-product-name-type-product-identification-helper-purls):
> The given pattern does not completely evaluate whether a PURL is valid according to the [[ECMA-427](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#ECMA-427)] specification. It provides a more generic approach and general guidance to enable forward compatibility. CSAF uses only the canonical form of PURL to conform with section 3.3 of [[RFC3986](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#RFC3986)]. Therefore, URLs starting with pkg:// are considered invalid.

The list of modifications to how CSAF does PURL could be extended on in the future, and PURL / the PURL test suite are also under development, so I see value in continously comparing the purl spec suite against how our reference implementation of CSAF acts.

csaf-rs/tests/purl_spec_conformance.rs
```diff
@@ -0,0 +1,167 @@
...
+        let mut paths: Vec<_> = std::fs::read_dir(&dir_path)
+            .unwrap_or_else(|e| {
+                panic!("{dir_path}: {e} — initialize the purl-spec submodule (git submodule update --init)")
+            })
+            .map(|entry| entry.expect("directory entry").path())
+            .collect();
+        paths.sort();
+        for path in paths {
+            let text = std::fs::read_to_string(&path).expect("suite file reads");
+            let suite: SuiteFile = serde_json::from_str(&text).expect("suite file parses");
```
peinjoh, line 60: Here (or above in the path section) should be a filter to json files. If the directory were to include a README.md (or anything non-json) this would break I think

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = csaf
 	url = https://github.com/oasis-tcs/csaf
 	branch = master
+[submodule "purl-spec"]
+	path = purl-spec
+	url = https://github.com/package-url/purl-spec.git
```
peinjoh, line 7: Minor: I would argue we should specify main here, for the small chance that the spec repo unexpectedly changes their default branch

csaf-rs/tests/purl_spec_conformance.rs
peinjoh: Thanks for the fix!

.gitmodules
peinjoh: Thanks for the fix!

gronke:
> I fully agree that the effort should be on packageurl-rs being improved, but I would also be interested in your implementation :)

https://github.com/scm-rs/packageurl.rs/pull/33 should get packageurl compliant with purl-spec. The other PR is updated to vendor the purl-spec tests, which I personally would prefer sourced from a remote repo. Not my call to make though. Let's see.

gronke:
Rebased onto main and reworked per your points 2 and 3: the suite files flow through update_assets.sh into `csaf-rs/assets/purl-spec/` like the csaf submodule's assets, and the known gaps are pinned in `tests/purl_spec_known_gaps.json`, keyed by CSAF version.
Dependabot plus `asset-sync.yml` make purl-spec bumps ordinary reviewable PRs, and the self-pruning file surfaces any behavioral drift they bring.

gronke:
The best move here appears to be waiting for the two PRs to land on packageurl. https://github.com/scm-rs/packageurl.rs/pull/32 will test for purl-spec compliance and https://github.com/scm-rs/packageurl.rs/pull/33 should establish parity, with or without the actual tests against the spec. If both lands, this PR can be closed. Turned it into a draft to indicate the waiting state.

gronke:
Both landed upstream, so that there is little need to test purl-spec compliance here. Will keep this open until packageurl v0.7.0 is released and bump the version instead.

gronke:
Keeping this open for confirmation of the remaining gaps. Feel free to close when no longer needed.

gronke:
I don't think there is reason to keep this open after package-url became fully compliant with purl-spec. Can be re-opened when needed for reference.

PR #713, opened 2026-07-18
build: packageurl requirement 0.6

gronke:
The pre-release requirement string resolves to the stable 0.6.0 either way; the caret range states it directly and stops advertising a pre-release in the published metadata.


peinjoh:
Thanks for the PR! The pre-release version was a leftover from us integrating 0.6 before the release, we should have changed that a long time ago, as 0.6 has been out for quite some time. 👍 

Based on the changelog and changeset, this only includes a bugfix and the chores for version release: https://github.com/scm-rs/packageurl.rs/releases/tag/v0.6.0

LGTM, merging.

PR #714, opened 2026-07-18
build: drop the docs.rs rustc-args cfg

gronke:
rustc-args applies --cfg docsrs to every dependency build; generic-array 0.14 reacts with a nightly feature removed in Rust 1.92 and the whole docs.rs build fails (E0557). Nothing in this crate reads the cfg.


peinjoh:
Thanks again for the PR! This flag was introduced here: https://github.com/csaf-rs/csaf/pull/566, in an effort to fix broken docsrs generation. We have since moved most of code relevant this to a seperate type generation crate, and the flag should have been removed alongside this code.

LGTM, merging.

PR #715, opened 2026-07-21
feat: typed CVSS accessors on ContentTrait

gronke:
The CVSS payloads on `Score` (CSAF 2.0) and `Content` (CSAF 2.1) are raw `serde_json::Map`, so a consumer reading scores works stringly-typed (`.get("vectorString")`, `.get("baseScore")`) or rebuilds the typed objects by hand — while the crate already deserializes the same maps into the cvss-rs types internally (`deserialize_cvss`, tests 6.1.09/6.1.10).

`get_cvss_v2_typed`, `get_cvss_v3_typed`, and `get_cvss_v4_typed` on `ContentTrait` parse the raw maps into `cvss_rs::v2_0::CvssV2` / `v3::CvssV3` / `v4_0::CvssV4`, following the existing `get_ssvc_v2` pattern (which already returns the external `ssvc` type the same way).
`None` when the metric is absent, `Some(Err(_))` when the map does not deserialize.

Note: this makes cvss-rs part of the public API surface, as ssvc already is through `get_ssvc_v2`.
An in-crate wrapper under `csaf/types/` would avoid that, at the cost of mirroring the cvss-rs structs — happy to rework in that direction if preferred.

### DIFF
--- a/csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
+++ b/csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
@@ -4,6 +4,7 @@ use crate::schema::csaf2_1::schema::{Content, Epss, QualitativeSeverityRating};
 use cvss_rs::v2_0::CvssV2;
 use cvss_rs::v3::CvssV3;
 use cvss_rs::v4_0::CvssV4;
+use serde::Deserialize;
 use serde::de::Error as SerdeError;
 use serde_json::{Map, Value};
 use ssvc::selection_list::SelectionList;
@@ -66,8 +67,7 @@ pub trait ContentTrait {
 
     /// Returns the contained CVSS 2.0 metric parsed into its typed representation, if any.
     fn get_cvss_v2_typed(&self) -> Option<Result<CvssV2, serde_json::Error>> {
-        self.get_cvss_v2()
-            .map(|map| serde_json::from_value(Value::Object(map.clone())))
+        self.get_cvss_v2().map(CvssV2::deserialize)
     }
 
     /// Returns a JSON representation of the contained CVSS 3.0/3.1 metric, if any.
@@ -80,8 +80,7 @@ pub trait ContentTrait {
 
     /// Returns the contained CVSS 3.0/3.1 metric parsed into its typed representation, if any.
     fn get_cvss_v3_typed(&self) -> Option<Result<CvssV3, serde_json::Error>> {
-        self.get_cvss_v3()
-            .map(|map| serde_json::from_value(Value::Object(map.clone())))
+        self.get_cvss_v3().map(CvssV3::deserialize)
     }
 
     /// Returns a JSON representation of the contained CVSS 4.0 metric, if any.
@@ -94,8 +93,7 @@ pub trait ContentTrait {
 
     /// Returns the contained CVSS 4.0 metric parsed into its typed representation, if any.
     fn get_cvss_v4_typed(&self) -> Option<Result<CvssV4, serde_json::Error>> {
-        self.get_cvss_v4()
-            .map(|map| serde_json::from_value(Value::Object(map.clone())))
+        self.get_cvss_v4().map(CvssV4::deserialize)
     }
 
     /// Returns whether this content contains any CVSS metric (v2, v3, or v4).
--- a/csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
+++ b/csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
@@ -248,50 +248,109 @@ impl ContentTrait for Content {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rstest::rstest;
     use serde_json::json;
 
-    #[test]
-    fn score_cvss_v3_typed_parses_the_map() {
-        let score: Score = serde_json::from_value(json!({
-            "products": ["CSAFPID-0001"],
-            "cvss_v3": {
-                "version": "3.1",
-                "vectorString": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H",
-                "baseScore": 9.8,
-                "baseSeverity": "CRITICAL"
-            }
-        }))
-        .expect("score deserializes");
-        let cvss = score.get_cvss_v3_typed().expect("present").expect("parses");
-        assert_eq!(cvss.vector_string, "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H");
-        assert!(score.get_cvss_v2_typed().is_none());
+    /// Builds a CSAF 2.1 `Content` carrying `metric` under `key`.
+    fn content_with(key: &str, metric: &Value) -> Content {
+        let mut map = Map::new();
+        map.insert(key.to_string(), metric.clone());
+        serde_json::from_value(Value::Object(map)).expect("content deserializes")
     }
 
-    #[test]
-    fn score_cvss_v3_typed_reports_a_nonconforming_map() {
-        let score: Score = serde_json::from_value(json!({
-            "products": ["CSAFPID-0001"],
-            "cvss_v3": { "version": "3.1" }
-        }))
-        .expect("score deserializes");
-        assert!(score.get_cvss_v3_typed().expect("present").is_err());
+    /// Builds a CSAF 2.0 `Score` carrying `metric` under `key`.
+    fn score_with(key: &str, metric: &Value) -> Score {
+        let mut map = Map::new();
+        map.insert("products".to_string(), json!(["CSAFPID-0001"]));
+        map.insert(key.to_string(), metric.clone());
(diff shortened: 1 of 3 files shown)


csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
peinjoh, line 55: Site note: Your signature return type `Option<Result<...>` is better than the one we have in get_ssvc_v2 so far, as it can distinguish between property not preset and actual serde_json parsing errors.

I would be in favor of refactoring get_ssvc_v2 to match your approach, but this is outside of the scope of this PR.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
```diff
@@ -223,3 +244,54 @@ impl ContentTrait for Content {
...
+            }
+        }))
+        .expect("content deserializes");
+        let cvss = content.get_cvss_v4_typed().expect("present").expect("parses");
+        assert_eq!(
+            cvss.vector_string,
+            "CVSS:4.0/AV:N/AC:L/AT:N/PR:N/UI:N/VC:H/VI:H/VA:H/SC:N/SI:N/SA:N"
+        );
+    }
+}
```
peinjoh: This might be a bit of overtesting, as all 3 functions work very similar, but I would be in favor of seperating and completing the test coverage to avoid regression here.

Currently, the following is tested:
* v3 provided, v3 works, v2 is not present
* broken v3 provided, error is reported
* v4 provided, v4 works

There are 3 seperate test concerns here:
1) valid v2/3/4 is provided, and should work as intended when returned
2) broken v2/3/4 is provided, and should return an error
3) valid v2/3/4 is provided, and the other two should return none (v2 is provided -> v3,v4 return none)

IMO 1) and 3) can be merged, i.e. by checking that the returned value is working as intended and the other two cvss versions return none.

We use rstest elsewhere in this project, so this could be two rstest blocks with 3 cases each. 

I'm fine with merging as is / extending the test coverage myself if this adds too much work to this PR for your prefence 👍

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
```diff
@@ -69,6 +78,12 @@ pub trait ContentTrait {
         self.get_cvss_v3().is_some()
     }
 
+    /// Returns the contained CVSS 3.0/3.1 metric parsed into its typed representation, if any.
+    fn get_cvss_v3_typed(&self) -> Option<Result<CvssV3, serde_json::Error>> {
+        self.get_cvss_v3()
+            .map(|map| serde_json::from_value(Value::Object(map.clone())))
```
peinjoh: Minor: cvss maps are **relativly** small, but I think there can be a performance impact of doing a lot cloning here, specially in bigger documents with lots of cvss score. If this can be easily done by ref, we should do it, but otherwise fine as-is.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
gronke: 🗒️ Noted

Remind me after this PR in case I forget about getting back to it.

peinjoh:
> Note: this makes cvss-rs part of the public API surface, as ssvc already is through get_ssvc_v2.
An in-crate wrapper under csaf/types/ would avoid that, at the cost of mirroring the cvss-rs structs — happy to rework in that direction if preferred.

I don't think mirroring is necessary, as I expect cvss-rs to be stable and "done" before this project reaches that stage. Otherwise, we can still mirror later.

> get_cvss_v2_typed, get_cvss_v3_typed, and get_cvss_v4_typed on ContentTrait parse the raw maps into cvss_rs::v2_0::CvssV2 / v3::CvssV3 / v4_0::CvssV4, following the existing get_ssvc_v2 pattern (which already returns the external ssvc type the same way).
None when the metric is absent, Some(Err(_)) when the map does not deserialize.

Your approach is slightly different from how ssvc_v2 works (as described above), but I like it much better, thank you for that!

Also, thank you for this improvement in general! I would lastly like to know what motivated the changes, i.e. if you have a use case / use cases that require this and that we should consider in our further development of the library. 👍

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
```diff
@@ -1,6 +1,9 @@
 use crate::csaf::types::csaf_vuln_metric::CsafVulnerabilityMetric;
 use crate::schema::csaf2_0::schema::Score;
 use crate::schema::csaf2_1::schema::{Content, Epss, QualitativeSeverityRating};
+use cvss_rs::v2_0::CvssV2;
+use cvss_rs::v3::CvssV3;
+use cvss_rs::v4_0::CvssV4;
```
peinjoh, line 6: As we now expose the cvss structs, we should a) document that users will need cvss-rs in their cargo toml or b) re-export the structs. The same applies to ssvc I think! 

(This is also outside the scope of this PR and should be investigated and fixed seperatly.)

gronke:
> Your approach is slightly different from how ssvc_v2 works (as described above), but I like it much better, thank you for that!
> 
> Also, thank you for this improvement in general! I would lastly like to know what motivated the changes, i.e. if you have a use case / use cases that require this and that we should consider in our further development of the library. 👍

Thanks, I'm glad to hear you appreciate the contribution. My original motivation started with an [audit feature in `npm-utils`](https://github.com/gronke/npm-utils/tree/main/src/audit) which raised  need for a downstream library that normalizes security advisories from several formats (CSAF among them) into one neutral audit model with typed multi-version CVSS ratings.
Ingesting CSAF today means reading `baseScore`/`vectorString` stringly out of the raw metric maps and rebuilding the typed objects by hand; the typed accessors delete that shim.

#716 and #717 come from the same integration: resolving `product_status` references means walking the product tree by hand, and comparing generated string newtypes against plain strings needs `.0` unwrapping.
Beyond those, the two themes that would help this kind of consumer most: a models-only build (parse and serialize without the validation machinery and its dependency tree), and a lenient parsing mode for reading documents as they exist in the wild - #693 already discusses one shape of that.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
gronke: It can. serde_json implements `Deserializer` for `&Map<String, Value>` since 1.0.131 (https://github.com/serde-rs/json/pull/1135), so the accessors now deserialize straight from the borrowed map instead of cloning it: self.get_cvss_v3().map(CvssV3::deserialize) (a5d2d61, likewise for v2/v4). I've set the serde_json floor to 1.0.135 in Cargo.toml; the lockfile already resolves 1.0.149, so nothing else shifts.
I've set the serde_json floor to 1.0.131 in Cargo.toml; the lockfile already resolves 1.0.149, so nothing else shifts.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
gronke: I went with option (b) and re-export the structs.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
peinjoh: You can also just open an issue for this if you don't feel like doing this specific task, although we would be thankful for the refactoring!

This might also be candidate for the good first issue tag :)

gronke: Good call, please find https://github.com/csaf-rs/csaf/issues/736 as a note and feel free to tag it as good first issue. No rush in anything here.

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
peinjoh: Looks good 👍

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
peinjoh: Perfect. Taking a note, this might allow a perf improvement in other places, too 👍 (#742)

csaf-rs/src/csaf/traits/vulnerabilities/content_trait.rs
peinjoh: Perfect, resolved!

peinjoh (approved):
LGTM 🚀

gronke:
@peinjoh want me to rebase once more? I've enabled edits by maintainers. During ongoing change that feature tends to confuse me, but I'll leave the final rebase up to you.

peinjoh:
I've done the rebase, feel free to press the button when the checks are done 👍

gronke:
I don't have permission to merge, but thanks for the offer.

peinjoh:
Taking notes, this might need a bit of tinkering with the github config 👀

PR #716, opened 2026-07-21
feat: product lookup by ID on ProductTreeTrait

gronke:
Resolving a `product_id` reference (from `product_status`, flags, threats, or scores) to its product definition has no library support: the visitors yield `(product, json_pointer)` pairs shaped for validation reporting, and `get_products_by_status` returns owned strings, so a consumer ends up hand-walking branches, `full_product_names`, and relationships.

`ProductTreeTrait::get_product_by_id` returns the product definition carrying a given product ID, searching the same three sources as `visit_all_products_generic`; `BranchTrait::find_product_by_id` carries the recursive half.
Both are default methods on the existing traits, and the returned reference borrows from the tree, so callers keep working with the typed product.

Tests cover branch-nested, top-level, and relationship products for both CSAF 2.0 and 2.1.

### DIFF
--- a/csaf-rs/src/csaf/traits/product_tree_trait.rs
+++ b/csaf-rs/src/csaf/traits/product_tree_trait.rs
@@ -91,26 +91,31 @@ pub trait ProductTreeTrait {
 
     // Lookup functions
 
-    /// Returns the product definition carrying the given product ID, searching the same
+    /// Returns every product definition carrying the given product ID, searching the same
     /// sources as `visit_all_products_generic()`: branches (recursively), the top-level
     /// full product names, and (for 2.0) the relationships' full product names, or (for 2.1) the product-paths' full product names.
-    fn get_product_by_id(&self, product_id: &str) -> Option<&Self::FullProductNameType> {
-        if let Some(branches) = self.get_branches()
-            && let Some(product) = branches.iter().find_map(|branch| branch.find_product_by_id(product_id))
-        {
-            return Some(product);
+    ///
+    /// A valid document defines each product ID once (test 6.1.2), so any element beyond the
+    /// first is a duplicate definition; handling those is the caller's choice.
+    fn get_products_by_id(&self, product_id: &str) -> Vec<&Self::FullProductNameType> {
+        let mut products = Vec::new();
+        if let Some(branches) = self.get_branches() {
+            for branch in branches {
+                branch.collect_products_by_id(product_id, &mut products);
+            }
         }
-        if let Some(product) = self
-            .get_full_product_names()
-            .iter()
-            .find(|fpn| fpn.get_product_id() == product_id)
-        {
-            return Some(product);
-        }
-        self.get_product_paths()
-            .iter()
-            .map(|path| path.get_full_product_name())
-            .find(|fpn| fpn.get_product_id() == product_id)
+        products.extend(
+            self.get_full_product_names()
+                .iter()
+                .filter(|fpn| fpn.get_product_id() == product_id),
+        );
+        products.extend(
+            self.get_product_paths()
+                .iter()
+                .map(|path| path.get_full_product_name())
+                .filter(|fpn| fpn.get_product_id() == product_id),
+        );
+        products
     }
 
     // Visitors for node types in the tree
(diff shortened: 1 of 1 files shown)


csaf-rs/src/csaf/traits/product_tree_trait.rs
```diff
@@ -89,6 +89,30 @@ pub trait ProductTreeTrait {
         ids
     }
 
+    // Lookup functions
+
+    /// Returns the product definition carrying the given product ID, searching the same
+    /// sources as `visit_all_products_generic()`: branches (recursively), the top-level
+    /// full product names, and the relationships' full product names.
```
peinjoh: Minor: Word is slightly incorrect for CSAF 2.1, which has product_paths instead of relationships.
```suggestion
    /// full product names, and (for 2.0) the relationships' full product names, or (for 2.1) the product-paths' full product names.
```

csaf-rs/src/csaf/traits/product_tree_trait.rs
```diff
@@ -428,3 +465,72 @@ impl BranchTrait<FullProductNameT21> for Branch21 {
...
+        }))
+        .expect("product tree deserializes");
+
+        let branch_product = tree.get_product_by_id("CSAFPID-0001").expect("branch product");
+        assert_eq!(branch_product.get_name(), "Vendor Product 1.0");
+        let top_level = tree.get_product_by_id("CSAFPID-0002").expect("top-level product");
+        assert_eq!(top_level.get_name(), "Top-level product");
+        let relationship = tree.get_product_by_id("CSAFPID-0003").expect("relationship product");
+        assert_eq!(relationship.get_name(), "Product on top-level");
+        assert!(tree.get_product_by_id("CSAFPID-9999").is_none());
```
peinjoh: Question: What motivates the asymetry in the testing behaviour between CSAF 2.0 and 2.1, i.e. why is CSAF 2.0 being tested via expect / get_name and CSAF 2.1 via "just" is_some?

csaf-rs/src/csaf/traits/product_tree_trait.rs
```diff
@@ -89,6 +89,30 @@ pub trait ProductTreeTrait {
         ids
     }
 
+    // Lookup functions
+
+    /// Returns the product definition carrying the given product ID, searching the same
+    /// sources as `visit_all_products_generic()`: branches (recursively), the top-level
+    /// full product names, and the relationships' full product names.
+    fn get_product_by_id(&self, product_id: &str) -> Option<&Self::FullProductNameType> {
+        if let Some(branches) = self.get_branches()
```
peinjoh: I agree that this function helpful, but the return signature is conceptually misleading. Your function assumes that product IDs are unique, and returns the first occurance. But this assumption is not always true, as evidenced by the standard containing test "6.1.2 Multiple Definition of Product ID" to validate documents against this (and some schema uniqueness checks). 

So the return value of this function should IMO be a collection of sorts, leaving it to the user how to handle the returned FPNs if the product ID is not actually unique across the document.

csaf-rs/src/csaf/traits/product_tree_trait.rs
gronke: No motivation, just an inconsistency. Thanks for pointing it out!

`test: assert the product names for every 2.1 lookup source` aligns the 2.1 expectations with the 2.0 test, and the commit following it keeps the name assertions for every source.

csaf-rs/src/csaf/traits/product_tree_trait.rs
gronke: Agreed. 6.1.2 exists precisely because parsed documents can define an ID twice. `feat: collect every product definition for a product ID` renames the lookup to `get_products_by_id` returning `Vec<&FullProductNameType>` with every match in traversal order (branches, full product names, relationships / product paths), so duplicate handling stays with the caller. Both CSAF versions now have a test that a doubly-defined ID yields both definitions in order.

peinjoh: Perfect, thank you for that improvement!

peinjoh (approved):
LGTM, @gronke if there is nothing left from your side to do here, this can be merged! 👍

gronke:
Rebased on the default branch for good measure. Confirm, ready to merge from my side :+1:

PR #717, opened 2026-07-21
feat: AsRef<str> and PartialEq<str> for generated string newtypes

gronke:
The generated constrained-string newtypes (`ProductIdT`, `VersionT`, …) offer `Deref<Target = String>` but no `AsRef<str>` or `PartialEq<str>`, so comparison sites read like `**a == **b` and callers dereference through the inner `String` for lookups and `.contains` checks.

The type-generator now appends `impl AsRef<str>`, `impl PartialEq<str>`, and `impl PartialEq<&str>` for every generated single-field `String` tuple struct, and the regenerated schema files carry the impls (the diff is purely additive).
Detection is shape-based in a `syn::File` post-pass next to the existing file-attribute passes, so future newtypes pick the impls up automatically.

An integration test pins the ergonomics for `ProductIdT` in both schema versions.


peinjoh:
Will look into this next 👍

gronke:
Maintainer edits granted, in case just a rebase is needed. I'll be afk for a bit. 👋

type-generator/src/utils/codegen_snippets.rs
```diff
@@ -28,3 +28,57 @@ pub fn add_generated_code_header(file: &mut syn::File) {
...
+            }
+        });
+        impls.push(syn::parse_quote! {
+            impl ::std::cmp::PartialEq<&str> for #ident {
+                fn eq(&self, other: &&str) -> bool {
+                    self.0 == *other
+                }
+            }
+        });
+    }
```
peinjoh, line 52: Great improvement to current ergonomics, but this could be even better if we make it cummutative / make it work for reversed param order.

Currently `assert_eq!(id, "CSAFPID-0001")` works, but `assert_eq!("CSAFPID-0001", id)` breaks the build. Should be straightforward with two additional blocks, though. 👍

peinjoh, line 52: Maybe even 3 blocks for PartialEq<String> to avoid the deref there? I believe there is some wierdness when making this one cummutative, but would be happy to have a look at this one 👀

peinjoh (changes requested):

peinjoh:
> Maintainer edits granted, in case just a rebase is needed. I'll be afk for a bit. 👋

Sure, I'll pick this up then. Have a great time afk 👍

type-generator/src/utils/codegen_snippets.rs
gronke: The impls now live in one `impl_string_newtype_ergonomics!` macro (the `define_csaf_test!` pattern), so this and future additions no longer bloat the generated files.
The ergonomics test pins your example: `assert_eq!("CSAFPID-0001", id)`.

gronke: The weirdness is real, but it only affects the generic formulations, for different coherence reasons. `impl<T: AsRef<str>> PartialEq<T> for String` hits E0210: neither the trait nor `String` is local, and `T` is uncovered. The local-side generic form, `impl<T: AsRef<str>> PartialEq<T> for ProductIdT`, overlaps the derived `PartialEq` at `T = ProductIdT` because the macro also implements `AsRef<str>` for the newtype (E0119). Concrete per-newtype impls avoid both; `impl PartialEq<ProductIdT> for String` is coherence-legal because `ProductIdT` is a local type participating in the impl.

`f845ae6` adds the owned `String` pair in both operand orders, completing the `str`/`&str`/`String` value matrix. If “3 blocks” was also meant to cover the borrowed-`String` forms (`ProductIdT == &String` and `&String == ProductIdT`), those are not included yet; I can add that pair as well. The current shape follows the cross-comparison style used by `smol_str::SmolStr` and `bytes::Bytes`, although `SmolStr` also spells out the `&String` pair.

Thanks for offering to double-check this one 👀

gronke:
It reads much better as a macro, for less scrolling during (the next) review. 😅 

@peinjoh thank you for the feedback. I've responded to the code comments directly, happy to adjust to your new input.

tziemek (approved):

peinjoh:
On it 👍

type-generator/src/utils/codegen_snippets.rs
peinjoh: Thanks, looking good!

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Generates:
+/// - `AsRef<str>` borrowing the inner `String`
+/// - `PartialEq<str>`, `PartialEq<&str>`, and `PartialEq<String>`, in both operand orders
+///
+/// Invocations of this macro are generated by the type-generator for every single-field
+/// `String` tuple struct in the `schema/*/schema.rs` and `schema/*/testcases_schema.rs` files.
+macro_rules! impl_string_newtype_ergonomics {
+    ($ident:ident) => {
+        impl ::std::convert::AsRef<str> for $ident {
```
peinjoh, line 12:
```suggestion
        // newtype.as_ref() returns &str without deref
        impl ::std::convert::AsRef<str> for $ident {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn as_ref(&self) -> &str {
+                &self.0
+            }
+        }
+
+        impl ::std::cmp::PartialEq<str> for $ident {
```
peinjoh, line 19:
```suggestion
        // newtype == str without deref
        impl ::std::cmp::PartialEq<str> for $ident {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn eq(&self, other: &str) -> bool {
+                self.0 == other
+            }
+        }
+
+        impl ::std::cmp::PartialEq<&str> for $ident {
```
peinjoh, line 26:
```suggestion
        // newtype == &str without deref
        impl ::std::cmp::PartialEq<&str> for $ident {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn eq(&self, other: &&str) -> bool {
+                self.0 == *other
+            }
+        }
+
+        impl ::std::cmp::PartialEq<::std::string::String> for $ident {
```
peinjoh, line 33:
```suggestion
        // newtype == String without deref
        impl ::std::cmp::PartialEq<::std::string::String> for $ident {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn eq(&self, other: &::std::string::String) -> bool {
+                self.0 == *other
+            }
+        }
+
+        impl ::std::cmp::PartialEq<$ident> for str {
```
peinjoh, line 40:
```suggestion
        // str == newtype without deref
        impl ::std::cmp::PartialEq<$ident> for str {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn eq(&self, other: &$ident) -> bool {
+                *self == other.0
+            }
+        }
+
+        impl ::std::cmp::PartialEq<$ident> for &str {
```
peinjoh, line 47:
```suggestion
        // &str == newtype without deref
        impl ::std::cmp::PartialEq<$ident> for &str {
```

csaf-rs/src/macros/string_newtype.rs
```diff
@@ -0,0 +1,55 @@
...
+            fn eq(&self, other: &$ident) -> bool {
+                **self == other.0
+            }
+        }
+
+        impl ::std::cmp::PartialEq<$ident> for ::std::string::String {
```
peinjoh, line 54:
```suggestion
        // String == newtype without deref
        impl ::std::cmp::PartialEq<$ident> for ::std::string::String {
```

type-generator/src/utils/codegen_snippets.rs
peinjoh: Thank you, also for the great explaination! 

This is exactly what I meant by 3 blocks. From our side, `&String` is not necessary, if we ever see the need to add it / are requested to add it, we'll have an easy time adding it now 👍

peinjoh (approved):
Looking good, I added a suggestion per macro'ed impl block to clear up to first contributors what each block does, otherwise thank you for this improvement to our libs ergonomics 💯

peinjoh:
If it's fine with you, I'll apply my suggestions and then merge?

gronke:
Sounds good, I was wondering how to merge the changes without n different commits.

peinjoh:
> Sounds good, I was wondering how to merge the changes without n different commits.

It thought the "apply in batch" functionality in the code review view was available to PR owners, too. Is it not? 👀

Issue #742, opened 2026-07-23
Research: perf improvement through serde_json deserializer for refs


peinjoh:
From https://github.com/csaf-rs/csaf/pull/715#discussion_r3628904434

> It can. serde_json implements Deserializer for &Map<String, Value> since 1.0.131 (https://github.com/serde-rs/json/pull/1135), so the accessors now deserialize straight from the borrowed map instead of cloning it: self.get_cvss_v3().map(CvssV3::deserialize) (https://github.com/csaf-rs/csaf/commit/a5d2d610c66bcede8c98ce99ea3a4eb8b32b4ca0, likewise for v2/v4). I've set the serde_json floor to 1.0.135 in Cargo.toml; the lockfile already resolves 1.0.149, so nothing else shifts.
I've set the serde_json floor to 1.0.131 in Cargo.toml; the lockfile already resolves 1.0.149, so nothing else shifts.

Research task: This might allow a perf improvement in other places, too

gronke:
Surveyed the crate for more of these. Three sites qualified; everything else either consumes an owned `Value` legitimately (`JsonSource for Value`, the type-generator's schema loading) or feeds jsonschema, which needs a concrete `&Value`, so the clone there is structural rather than a deserialization.

- `RawDocument::get_parsed` cloned the entire document `Value` before `from_value` - paid once per document on every validation and FFI path. Now `T::deserialize(&self.raw)`.
- `cvss::deserialize_cvss` cloned each metric map into a `Value::Object` per call. Now `Cvss::deserialize(cvss_map)`.
- `get_ssvc_v2` cloned the ssvc map per call. Now `SelectionList::deserialize(&self.ssvc_v2)` - body only, the signature alignment stays with #736.

gronke:
Opened two pull-requests, the second of which emerged while investigating lead from PR #715 . I did not measure performance on the combined effort. Since we add `bench_typed_parse` in #747 the results to previous benchmarks are not directly comparable unfortunately. @peinjoh Let me know if you would prefer to split this out in a separate PR to keep a realistic trail of the performance effects visible.

Issue #736, opened 2026-07-22
Align get_ssvc_v2 with the typed accessors' Option<Result<..>> signature


gronke:
Follow-up from a review discussion in #715 (https://github.com/csaf-rs/csaf/pull/715#discussion_r3622972075): the typed CVSS accessors distinguish "metric absent" from "metric present but malformed" through their return type,

```rust
fn get_cvss_v3_typed(&self) -> Option<Result<CvssV3, serde_json::Error>>
```

while `get_ssvc_v2` predates that shape and folds both cases into `Err`:

```rust
fn get_ssvc_v2(&self) -> Result<SelectionList, serde_json::Error>
```

An absent SSVC metric is not an error condition, and the CSAF 2.0 implementation (where SSVC does not exist) has to fabricate one ("SSVC metrics are not implemented in CSAF 2.0").

Proposed change:

- `ContentTrait::get_ssvc_v2` returns `Option<Result<SelectionList, serde_json::Error>>` — `None` when the metric is absent, `Some(Err(_))` when it is present but does not deserialize.
- The CSAF 2.0 implementation returns `None` instead of the custom error.
- The CSAF 2.1 implementation returns `None` for an empty `ssvc_v2` map and `Some(serde_json::from_value(..))` otherwise.
- `has_ssvc_v2` can become a default method on the trait (`self.get_ssvc_v2_raw().is_some()`), mirroring `has_cvss_v2`/`has_cvss_v3`/`has_cvss_v4`.

Call sites to adjust, all in-crate: the two implementations in `src/csaf/traits/vulnerabilities/content_trait.rs`, and the `match content.get_ssvc_v2()` arms in `src/validations/test_6_1_47.rs`, `test_6_1_48.rs`, and `test_6_1_49.rs`.

This changes `ContentTrait`'s public signature, so it is a candidate to ride along with the next breaking release.

PR #747, opened 2026-07-24
perf: deserialize typed documents and metrics from borrowed JSON

gronke:
Addresses #742. Found three clone-then-deserialize sites, all in csaf-rs, all covered by the `Deserializer` impls for `&Value`/`&Map` that #715 already set the serde_json floor for.

- `RawDocument::get_parsed` cloned the entire document `Value` before `from_value`, once per document on every validation, FFI, and converter path — now `T::deserialize(&self.raw)`.
- `deserialize_cvss` cloned each CVSS metric map into a `Value::Object` per call — now `Cvss::deserialize(cvss_map)`; the map is reused after the call, so the borrow is exactly right.
- `Content::get_ssvc_v2` cloned the ssvc map per call — now `SelectionList::deserialize(&self.ssvc_v2)`, body only; the `Option<Result<…>>` signature refactor stays with upstream #736.

The first commit adds a `typed_parse` bench: every existing bench pre-parses documents outside `b.iter`, so the whole-document clone was invisible to the suite and its regression alert.

### DIFF
--- a/csaf-rs/benches/validation_benchmark.rs
+++ b/csaf-rs/benches/validation_benchmark.rs
@@ -5,6 +5,7 @@ use std::time::Duration;
 
 use criterion::{Criterion, criterion_group, criterion_main};
 
+use csaf::csaf::raw::HasParsed;
 use csaf::csaf2_0::loader::load_document as load_document_2_0;
 use csaf::csaf2_0::testcases::{
     informative_tests as informative_tests_2_0, mandatory_tests as mandatory_tests_2_0,
@@ -221,6 +222,33 @@ fn bench_parse_only(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmark parsing plus the typed deserialization (no validation).
+///
+/// `parse_only` stops at the JSON `Value`; the delta between the two groups isolates the
+/// cost of materializing the typed document that every validation run pays once per document.
+fn bench_typed_parse(c: &mut Criterion) {
+    let fixtures_dir = concat!(env!("CARGO_MANIFEST_DIR"), "/../csaf/csaf_2.0/test/validator/data");
+    let contents = load_fixture_contents(&collect_fixture_files(fixtures_dir));
+
+    if contents.is_empty() {
+        return;
+    }
+
+    let mut group = c.benchmark_group("typed_parse");
+
+    group.bench_function("csaf_2_0", |b| {
+        b.iter(|| {
+            for (_name, content) in &contents {
+                if let Ok(doc) = load_document_2_0(content) {
+                    let _ = black_box(doc.get_parsed());
+                }
+            }
+        });
+    });
+
+    group.finish();
+}
+
 fn configured_criterion() -> Criterion {
     Criterion::default()
         .warm_up_time(Duration::from_secs(1))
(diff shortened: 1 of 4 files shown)


gronke:
Local comparison between the two commits — the bench lands first, so `HEAD~1` runs the same bench against the pre-change code.

Medians from criterion, suite defaults (1 s warm-up, 1 s measurement, 10 samples):

| bench | `HEAD~1` | `HEAD` | change |
|---|---|---|---|
| `typed_parse/csaf_2_0` | 2.93 ms | 2.07 ms | **−27 %** (p < 0.05) |
| `csaf_2_0_tests/6.1.9` | 83.6 µs | 68.0 µs | −17 % (p < 0.05) |
| `csaf_2_0_tests/6.1.10` | 146.4 µs | 129.4 µs | −13 % (p < 0.05) |
| `csaf_2_1_tests/6.1.9` | 559.9 µs | 511.4 µs | −13 % (p < 0.05) |
| `csaf_2_1_tests/6.1.47` | 486.2 µs | 407.8 µs | −15 % (p < 0.05) |
| `parse_only/csaf_2_0` (control, code untouched) | 1.24 ms | 1.24 ms | n.s. |
| `full_validation` 2.0 / 2.1 | 8.60 / 58.7 ms | 8.75 / 58.8 ms | n.s. |

6.1.8, 6.1.48, 6.1.49, and 2.1's 6.1.10 stayed within noise (p > 0.05). Subtracting the control from `typed_parse`, the typed materialization itself drops from ~1.70 ms to ~0.84 ms over the 239 CSAF 2.0 fixtures — roughly halved. `full_validation` is flat by design: it parses its documents outside the timed loop, which is exactly the blind spot the new bench closes.

To repeat the comparison locally:

```sh
git fetch origin pull/747/head:pr-747
git checkout pr-747~1                # first commit: the bench alone, code unchanged
git submodule update --init csaf     # the fixture corpus the benches read
FILTER='(typed_parse|parse_only|full_validation|6\.1\.(8|9|10|47|48|49)$)'
cargo bench --bench validation_benchmark -- --save-baseline before "$FILTER"
git checkout pr-747                  # second commit: borrowed-JSON deserialization
cargo bench --bench validation_benchmark -- --baseline before "$FILTER"
```

The second run annotates every bench with the change against the saved baseline and its p-value. The filter limits the run to the affected benches plus the controls; drop it to run the whole suite. With the fast suite settings, single-digit percentages are noise — the p-values separate signal from it. These numbers are from a container, so treat them as directional; the benchmark workflow's tracked results remain the authoritative trend.

csaf-rs/benches/validation_benchmark.rs
```diff
@@ -221,6 +222,33 @@ fn bench_parse_only(c: &mut Criterion) {
...
+    let contents = load_fixture_contents(&collect_fixture_files(fixtures_dir));
+
+    if contents.is_empty() {
+        return;
+    }
+
+    let mut group = c.benchmark_group("typed_parse");
+
+    group.bench_function("csaf_2_0", |b| {
+        b.iter(|| {
```
peinjoh: Minor improvement: Would you mind adding a csaf_2_1 version for the typed_parse benchmark for completeness? I don't expect the results to be any different, but the type being parsed into is different, and I would like to highlight / integrate these into our benchmark CI 👍 

If you are busy otherwise, I'd be happy to make that change myself :)

peinjoh (approved):

peinjoh:
Hey there, that changeset looks good 👍 Thank your for that significant performance improvement and for highlighting that existing blind spot in our benchmarking process!

csaf-rs/benches/validation_benchmark.rs
gronke: Looking into this now.

peinjoh: Thanks, also for competing the symetry in the `bench_parse_only` function, although I expect even less of a difference there. The RawDocument should be very similar between versions, as the "typed" difference comes from parsing, for which we have an extra benchmark for now 👍

peinjoh:
Looking good, if you are done here, I'll merge 👍

gronke:
LGTM :+1:

PR #749, opened 2026-07-24
feat: extend cwe xslt with `Mapping_Notes/Usage` field

peinjoh:
This PR:
* extends the cwe xslt to add two fields to the generated csv
    * 3rd field: is Mapping_Notes/Usage available for this entry
    * 4th field: Mapping_Notes/Usage, empty if unavailable
* reruns the xslt for all cwes

The 4th field would have technically been enough, as we can expect the field to be filled if present, but this should make importing a bit quicker.

This should make implementation of 6.2.25/26 easier, as only extending the helpers.rs section and implementing the check is necessary.


peinjoh:
Side note: The xslt lacks input sanitization for all fields. Which is kind of fine, as cwe.org is a well-known, reputable source. Still, after we move this code to the cwe repo with type generation, input sanitization (whitespace normalization, tabs, CR, LF removal in values, ...) should be implemented.

peinjoh:
Rebase to accomodate the changes from #695 (impl of test 6.2.23, adding the status field)

tziemek (approved):

PR #752, opened 2026-07-24
perf: parse each document once on the version-detection path

gronke:
Implements the double-parse finding from the #742 survey.

Auto mode in csaf-validator and csaf-result-json read and parsed every document twice - `detect_version` and `load_document` each did the full file read plus string→`Value` parse - and csaf-ffi's `from_json` parsed the identical string twice (the path wasm and Go consumers take). The enabling API already existed: `detect_version_with` returns `VersionAndData` for exactly this reuse, and `JsonSource` is implemented for `serde_json::Value`, so the fix is wiring the call sites - no csaf-rs changes at all.

- Both CLIs gain a small `load_and_validate(source: impl JsonSource, version, …)` helper; auto mode feeds it `detected.data`, explicit mode the path as before (one parse either way now).
- The FFI keeps its public constructors unchanged and routes `from_json` through private `from_value_2_0/2_1` helpers that keep the original string for `to_json()`.

Measured end to end (release csaf-validator, auto mode, one cheap test over the 902-file CSAF 2.1 corpus ×10, local and directional): wall time 0.61 s → 0.48 s (−22%), user CPU −20%.
The share grows with document size and the fixture corpus averages ~1 KB, real advisories are much larger.

Out of scope, per the survey: caching the reference accessors (needs API design) and jsonschema schema-test visibility; the stale 6.1.14 bench skip is separate hygiene.

### DIFF
--- a/csaf-result-json/src/main.rs
+++ b/csaf-result-json/src/main.rs
@@ -173,9 +173,6 @@ fn main() -> Result<(), anyhow::Error> {
     }
 }
 
-/// Validate a CSAF document of the specified version with the provided arguments.
-///
-/// This prints the results of the tests on stdout.
 /// Load a document of the given version from any JSON source and validate it.
 fn load_and_validate<S: JsonSource>(source: S, version: &str, tests: &[&str]) -> Result<ValidationResult> {
     match version {
@@ -185,6 +182,9 @@ fn load_and_validate<S: JsonSource>(source: S, version: &str, tests: &[&str]) ->
     }
 }
 
+/// Validate a CSAF document of the specified version with the provided arguments.
+///
+/// This prints the results of the tests on stdout.
 fn validate_document<T>(document: T, version: &str, tests: &[&str]) -> ValidationResult
 where
     T: Validatable,
--- a/csaf-result-json/src/main.rs
+++ b/csaf-result-json/src/main.rs
@@ -8,9 +8,10 @@ use anyhow::{Result, bail};
 use clap::{Parser, Subcommand};
 use compare::compare_result_jsons;
 use convert::build_testresult_json;
-use csaf::csaf::loader::detect_version;
+use csaf::csaf::loader::detect_version_with;
 use csaf::csaf2_0::loader::load_document as load_document_2_0;
 use csaf::csaf2_1::loader::load_document as load_document_2_1;
+use csaf::json::JsonSource;
 use csaf::validation::{Validatable, ValidationResult, validate_by_tests};
 use result::ResultJson;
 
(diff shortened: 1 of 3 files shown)


peinjoh:
On it 👍

csaf-result-json/src/main.rs
```diff
@@ -193,6 +176,15 @@ fn main() -> Result<(), anyhow::Error> {
 /// Validate a CSAF document of the specified version with the provided arguments.
 ///
 /// This prints the results of the tests on stdout.
+/// Load a document of the given version from any JSON source and validate it.
+fn load_and_validate<S: JsonSource>(source: S, version: &str, tests: &[&str]) -> Result<ValidationResult> {
+    match version {
+        "2.0" => Ok(validate_document(load_document_2_0(source)?, "2.0", tests)),
+        "2.1" => Ok(validate_document(load_document_2_1(source)?, "2.1", tests)),
+        _ => bail!(format!("Invalid CSAF version: {version}")),
```
peinjoh: Minor: bail! does formatting for you, so the inner format! is redundant

```suggestion
        _ => bail!("Invalid CSAF version: {version}"),
```

csaf-result-json/src/main.rs
```diff
@@ -193,6 +176,15 @@ fn main() -> Result<(), anyhow::Error> {
...
+    }
+}
+
 fn validate_document<T>(document: T, version: &str, tests: &[&str]) -> ValidationResult
```
peinjoh, line 188: Minor: Missing doc comment
In csaf-validator, the same comment was split between the two functions, I think this just needs a line to be moved.

peinjoh (approved):
Otherwise LGTM 🚀 Thank you again for improving our toolings performance!

I checked for the suggested bail!/format! fix repo-wide, those are the only two spots. I would just apply the fixes here.

Is there otherwise still something to do from you side here?

csaf-validator/src/main.rs
```diff
@@ -63,17 +64,24 @@ fn main() -> Result<(), anyhow::Error> {
...
-            let document = load_document_2_0(path)?;
+            let document = load_document_2_0(source)?;
             Ok(validate_document(document, "2.0", args))
         },
         "2.1" => {
-            let document = load_document_2_1(path)?;
+            let document = load_document_2_1(source)?;
             Ok(validate_document(document, "2.1", args))
         },
         _ => bail!(format!("Invalid CSAF version: {version}")),
```
peinjoh: Same as above
```suggestion
        _ => bail!("Invalid CSAF version: {version}"),
```

gronke:
Suggestions applied 🙋‍♂️ Once again, thanks for the review.

csaf-result-json/src/main.rs
peinjoh: Looking good 👍

peinjoh:
Perfect, thanks. If there is nothing left to do here, I'll merge 👍

peinjoh (approved):

gronke:
LGTM 🚢

Issue #769, opened 2026-07-29
refactor: align test method error initialization


peinjoh:
We usually use `Option<Vec<ValidationError>>` to avoid the allocation until an error is found / skip it on the happy path.

We should make sure all tests do this consistently.

PR #771, opened 2026-07-29
chore: improve 6.1.5, 8, 27.12, 27.19, 55 and 6.2.13, 20, 23, 24 error handling

peinjoh:
Use Option<Vec<ValidationError>> instead of Vec<ValidationError>

Also wrapped some static error messages in LazyLocks

part of #769

### DIFF
--- a/csaf-rs/src/validations/test_6_1_27_19.rs
+++ b/csaf-rs/src/validations/test_6_1_27_19.rs
@@ -61,8 +61,8 @@ pub fn test_6_1_27_19_reference_to_superseding_document(doc: &impl CsafTrait) ->
     // the document is not valid, even if there is also a reference with correct summary and correct category.
     // So the incorrect category has precedence over the missing reference. This way the error message is more specific and hints more
     // directly to the wrong instance.
-    if errors.clone().is_some_and(|e| !e.is_empty()) {
-        return Err(errors.unwrap());
+    if let Some(errs) = errors {
+        return Err(errs);
     }
 
     // completely missing reference with correct summary and category has second precedence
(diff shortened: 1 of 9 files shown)


csaf-rs/src/validations/test_6_1_27_19.rs
```diff
@@ -61,8 +61,8 @@ pub fn test_6_1_27_19_reference_to_superseding_document(doc: &impl CsafTrait) ->
     // the document is not valid, even if there is also a reference with correct summary and correct category.
     // So the incorrect category has precedence over the missing reference. This way the error message is more specific and hints more
     // directly to the wrong instance.
-    if errors.clone().is_some_and(|e| !e.is_empty()) {
```
peinjoh, line 64: Note: the is_empty check for defensiveness is redundant here IMO. errors is only initialized to `Some` during the insertion of an error.

csaf-rs/src/validations/test_6_2_13.rs
```diff
@@ -5,20 +5,20 @@ use serde_json::Value;
...
-        true => Ok(()),
-        false => Err(errors),
-    }
+    let mut errors: Option<Vec<ValidationError>> = None;
+    check_sorted_recursive(json, "", &mut errors);
+    errors.map_or(Ok(()), Err)
 }
 
-fn check_sorted_recursive(value: &Value, path: &str, errors: &mut Vec<ValidationError>) -> bool {
+fn check_sorted_recursive(value: &Value, path: &str, errors: &mut Option<Vec<ValidationError>>) -> bool {
```
aschierl-xitaso: Do we still need the returned bool if we then only look if the errors option is set?

peinjoh: We don't, will fix

peinjoh: Simplified the validation logic code, if we don't need to aggregate that bool, we don't necessarily need the map/filter/all syntax

tziemek (approved):

PR #781, opened 2026-07-30
feat: add test 6.2.25 / 6.2.26

### LINKED ISSUE #311: Implement 6.2.25 Usage of CWE Not Allowed for Vulnerability Mapping for CSAF 2.1

### LINKED ISSUE #312: Implement 6.2.26 Usage of CWE Allowed with Review for Vulnerability Mapping for CSAF 2.1

peinjoh:
This PR:
* builds upon #749, which added extraction of the Usage field
* refactors helpers.rs CWE_ENTRIES to accomodate the fields
* adds stronger typing for CWE_ENTRIES inner types, which outgrew the tuples typing they previously had
* moves 6.1.11 and 6.2.23 to this stronger typing 
* adds 6.2.25 and 6.2.26, which required the additional field

I see value and need in a refactor of the cwe-related functions. 6.1.11 and helpers.rs duplicate a function to get the latest version, and the 6.2.23,25,26 value comparisons could be folded into the data structures for reuseability. Further, the current data structure could be simplified, as 6.2.23 does not need the specific status, just a bool if it is "Deprecated", and the Usage could be an enum for cheaper comparisons. I think this is fine for now, the tests work as intended (I think), but we should tackle this during the migration the csaf-rs/cwe repo in the coming weeks.

Two open issue: 
I'm opening a ticket upstream (https://github.com/oasis-tcs/csaf/issues/1512) for the wording of 6.2.26. IMO, there is uncertainity how 6.2.26 should handle usage "Prohibited" and "Discouraged", i.e. if it should also generate errors, which would duplicate 6.2.25, or if it should skip them (which would make sense IMO, but is not explicit in the prose).

Also: The test coverage for both tests can be extended. 6.2.25 does not contain a case for "Allowed-with-Review", and implementation checking only "Allowed" would pass the suite. 6.2.26 does not contain a case for "Prohibited" and "Discouraged", leading to uncertainty mentioned above.

Resolves #311 and resolves #312


tziemek (approved):
Aside from copilot comments, LGTM

PR #783, opened 2026-07-30
feat: add test 6.2.47

peinjoh:
This PR:
* moves the "generate canonical filename" and "has canonical url" functions into the trait, including test coverage
* moves 6.2.11, which previously contained the functions, to use them
* adds test 6.2.47, which needed a similar function "get all canonical urls"

Test coverage looks fine IMO, but I would invite you to discuss edge cases if you find any during review

### DIFF
--- a/csaf-rs/src/csaf/traits/document_trait.rs
+++ b/csaf-rs/src/csaf/traits/document_trait.rs
@@ -12,11 +12,26 @@ use crate::schema::csaf2_0::schema::{
     Tracking as Tracking20,
 };
 use crate::schema::csaf2_1::schema::{
-    CsafVersion as CsafVersion21, DocumentLevelMetaData as DocumentLevelMetaData21, Note as Note21,
-    Publisher as Publisher21, Reference as Reference21, RulesForDocumentSharing as RulesForDocumentSharing21,
-    Tracking as Tracking21,
+    CategoryOfReference as CategoryOfReference21, CsafVersion as CsafVersion21,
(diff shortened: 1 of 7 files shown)


csaf-rs/src/csaf/traits/document_trait.rs
```diff
@@ -69,6 +84,20 @@ pub trait DocumentTrait {
...
+    /// Returns the canonical URLs from this document's references.
+    fn get_canonical_urls(&self) -> Vec<&str> {
+        let expected_filename = self.get_tracking().get_canonical_filename();
+        canonical_url_candidates(self.get_references(), expected_filename.as_str()).collect()
+    }
+
+    /// Returns `true` if the document has at least one canonical URL.
+    fn has_canonical_url(&self) -> bool {
+        let expected_filename = self.get_tracking().get_canonical_filename();
+        canonical_url_candidates(self.get_references(), expected_filename.as_str())
```
tziemek, line 96: Why not use `get_canonical_urls` here? Allocation of Vec?

tziemek (approved):

csaf-rs/src/csaf/traits/document_trait.rs
peinjoh: Yes, get_canonical_urls will collect the Iterator into a Vec and for that runs each element through the 2 filters.

has_canonical_urls will only run until the first element matches both filters, so ideally less then all elements are evaled, and will not alloc a Vec

Issue #763, opened 2026-07-28
Improve benchmarking


peinjoh:
We should explore: 
* alternatives to criterion (divan?)
* alternatives to running our benchmarks on github runners (codspeed, bencher?)

tziemek:
@gronner Do you have any suggestions for benchmarking?

PR #784, opened 2026-08-01
ci: compare PR benchmarks against the merge base on the same runner

gronke:
The PR summary previously compared against the last gh-pages entry, measured on a different runner instance; the stored history swings 1.16-1.47x between main runs, so the table flagged noise as regressions. PRs now bench the merge base first and compare within one runner through an external data file; pushes to main keep storing the gh-pages trend unchanged. The PR job pays one extra compile-and-bench pass for it.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -34,20 +34,54 @@ jobs:
       - name: Update rustup
         run: rustup update stable --no-self-update
 
+      - name: Run benchmarks (merge base)
+        if: github.event_name == 'pull_request'
+        run: |
+          git fetch --no-tags --depth=1 origin ${{ github.event.pull_request.base.sha }}
+          git checkout --detach ${{ github.event.pull_request.base.sha }}
+          git submodule update --init
+          cargo bench --bench validation_benchmark -- --output-format bencher | tee base-output.txt
+          git checkout --detach ${{ github.sha }}
+          git submodule update --init
+
+      - name: Save the merge-base result as comparison baseline
+        if: github.event_name == 'pull_request'
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: CSAF Validation Benchmark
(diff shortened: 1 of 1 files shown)


tziemek:
Thanks for this suggestion. We already discussed improving the benchmarks internally and are currently thinking about using Bencher or CodSpeed for running benchmarks, which would make this PR obsolete. Do you have any experience with those platforms? Additionally we consider moving to Divan as it is easier to read, any suggestions?

gronke:
I don't have practical experience with either solution. Thanks for mentioning it, measuring cycles is an elegant approach to differentially measure if a PR got better or worse. Without any Cache warmup and measuring magic, that would already allow to compare a reference from the default branch with proposed changes on a PR, without needing to run the entire benchmark twice on the PR, in hopes the noise on the shared runner is somewhat consistent. Measuring time is hard, and taking out the complexity of attempting to do so has a clear advantage here - it would reduce benchmark duration/load/cost - compared to the approach proposed by this PR.

Divian looks neat and Bencher already has an open issue [`bencher/issues/321`](https://github.com/bencherdev/bencher/issues/321) requesting support for it. Looks very doable and would make a nice contribution. I'd prefer a solution that can be reproduced local (in an offline environment) - you could nicely ask contributors to run the benchmark against the published reference artifact from main before pushing to a PR.

If Divian if not a priority, [Gungraun](https://github.com/gungraun/gungraun) could be another alternative to consider, which would not require coordination/development like Bencher would here. In case you decide for Divian, you could possibly get started with CodSpeed relatively quickly and then attempt to reproduce the results with Bencher and contribute a PR with Divian support. Reliable results fast, freedom later?

@tziemek feel free to close this PR, it is obsolete with your new direction. I'll leave the branch up while it's open.

tziemek:
Thanks for you input, I'll have a look at Gungraun. 
Linking this to #763 for further discussions so we can close this one.

PR #793, opened 2026-08-05
use struct to receive expected results

### LINKED ISSUE #791: Change expected results to concrete type
The `expect` function in our test framework is not really readable. Convert from parameters to a struct, so each testcase can be identified easily.

tziemek:
resolves #791 and helps in identifying testcases more easily

### DIFF
--- a/csaf-rs/src/macros/test_gen.rs
+++ b/csaf-rs/src/macros/test_gen.rs
@@ -11,7 +11,7 @@
 /// # Usage
 /// ```ignore
 /// define_csaf_test!(
-///     Test6_1_1, ValidatorForTest6_1_1,
+///     Test6_1_1, ValidatorForTest6_1_1, ExpectedResults_6_1_1,
 ///     id: "6.1.1",
 ///     doc_type: crate::schema::csaf2_0::schema::CommonSecurityAdvisoryFramework,
 ///     version: "V2_0",
@@ -23,7 +23,7 @@
 /// ```
 macro_rules! define_csaf_test {
     (
-        $struct_name:ident, $validator_name:ident,
+        $struct_name:ident, $validator_name:ident, $expect_name:ident,
         id: $id:literal,
         doc_type: $doc_type:ty,
         version: $version:literal,
@@ -69,6 +69,14 @@ macro_rules! define_csaf_test {
             }
         }
 
+        /// Expected results for each test case of this test, keyed by case name.
+        #[cfg(test)]
+        #[derive(Debug)]
+        #[allow(non_camel_case_types)]
+        pub struct $expect_name {
+            $(pub $case_name: Result<(), Vec<crate::validation::ValidationError>>,)*
+        }
+
         #[cfg(test)]
         impl<
             V: crate::test_validation::TestValidator<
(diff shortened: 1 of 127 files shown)


peinjoh:
Do you have any hints how to efficiently review this?

csaf-rs/src/macros/test_gen.rs
```diff
@@ -11,7 +11,7 @@
 /// # Usage
 /// ```ignore
 /// define_csaf_test!(
-///     Test6_1_1, ValidatorForTest6_1_1,
+///     Test6_1_1, ValidatorForTest6_1_1, ExpectedResults_6_1_1,
```
aschierl-xitaso, line 14: Did you intentionally use a different naming scheme between `ValidatorForTest6_1_1` and `ExpectedResults_6_1_1`? (concerning the underscore before the first number)

tziemek: Yes, I found it easier to read, but didn't want to change the `ValidatorForTestX` in this PR.

csaf-rs/src/validations/test_6_1_04.rs
```diff
@@ -74,20 +76,20 @@ mod tests {
...
-        TESTS_2_1.test_6_1_4.expect(
-            case_threats,         // threats
-            case_vulnerabilities, // vulnerabilities
-            case_vuln_id,         // vulnerability ids
-            Ok(()),
-            Ok(()),
-            Ok(()),
-        );
+        TESTS_2_0.test_6_1_4.expect(ExpectedResults_2_0 {
+            case_01: case_threats.clone(), // threats
```
peinjoh, line 80: Minor: These comments can be removed.

tziemek:
> Do you have any hints how to efficiently review this?

This is basically chore. It wouldn't compile or the tests would fail if there were any errors.

csaf-rs/src/validations/test_6_1_13.rs
```diff
@@ -68,18 +70,18 @@ mod tests {
...
 
-        TESTS_2_0.test_6_1_13.expect(
-            case_01_missing_name("purl", ""),
-            case_02_or_s06_type_prohibits_namespace("purl", ""),
-            Ok(()),
-            Ok(()),
-        );
+        TESTS_2_0.test_6_1_13.expect(ExpectedResults_2_0 {
+            case_01: case_01_missing_name("purl", ""),
+            case_s06: case_02_or_s06_type_prohibits_namespace("purl", ""),
```
aschierl-xitaso: Why do we start with s06 here? (what happened to s01..s05?)

csaf-rs/src/validations/test_6_1_41.rs
```diff
@@ -62,19 +63,19 @@ mod tests {
         let nil_uuid_err = Err(vec![NIL_UUID_SHARING_GROUP_ERROR.clone()]);
 
         // Only CSAF 2.1 has this test with 6 test cases (4 error cases, 2 success cases)
-        TESTS_2_1.test_6_1_41.expect(
-            // Case 01: Max UUID without name
+        TESTS_2_1.test_6_1_41.expect(ExpectedResults {
+            case_01: // Case 01: Max UUID without name
```
aschierl-xitaso: mentioning "Case 01" in the comment seems redundant (also in a few more tests)

aschierl-xitaso (approved):

csaf-rs/src/validations/test_6_1_13.rs
tziemek: Because in testcases.json s01-s05 is missing. Good catch.

tziemek: Ah they are missing because we have those testfiles, but they fail schema validation and thus would result in OK test cases. We should reintroduce them when lenient parsing works, or drop them then.

csaf-rs/src/validations/test_6_1_41.rs
tziemek: Thank you, missed some of those as they were not written the same.

peinjoh (approved):

Issue #787, opened 2026-08-03
Improve build pipeline


tziemek:
Currently we have duplicated build pipeline code across the different repositories of this organization.
We should find a way to simplify some of these steps, like format check, linting, etc.

peinjoh:
@bshek-xitaso is working on this

bshek-xitaso:
This issue includes several sub-issues:

- https://github.com/csaf-rs/shared-workflows/issues/3
- https://github.com/csaf-rs/shared-workflows/issues/4
- https://github.com/csaf-rs/shared-workflows/issues/6
- https://github.com/csaf-rs/shared-workflows/issues/7
- https://github.com/csaf-rs/shared-workflows/issues/5
- https://github.com/csaf-rs/shared-workflows/issues/11
- https://github.com/csaf-rs/shared-workflows/issues/12
- https://github.com/csaf-rs/shared-workflows/issues/13

peinjoh:
These deduplication / unification effort will be tracked and organizated via the shared-workflow issues. Closing this PR to not imply some need to report all issues / PRs here.

PR #801, opened 2026-08-10
Use shared composite actions for Rust CI checks and build context

bshek-xitaso:
This PR addresses csaf-rs/csaf#787.

**Note:** Before merging this PR, please make sure that the dependent [shared-workflows PR](https://github.com/csaf-rs/shared-workflows/pull/1) has been merged first, since this workflow uses the composite actions introduced there.

This PR updates the `csaf` build workflow to use the shared composite actions from `csaf-rs/shared-workflows`

The existing repository-specific workflow logic remains in `csaf`. The `prepare` job keeps the same job outputs, so the downstream jobs can continue using the existing output references without changes.

The existing Clippy arguments and the position of the formatting, Clippy, and build-context logic in the workflow are preserved.

Implementation details and validation of the shared actions can be found in the dependent [shared-workflows PR](https://github.com/csaf-rs/shared-workflows/pull/1).


tziemek (approved):

bshek-xitaso:
This PR has not been merged yet. Should I work on https://github.com/csaf-rs/shared-workflows/issues/7 here? @peinjoh

bshek-xitaso:
Please merge this PR before merging https://github.com/csaf-rs/csaf/pull/828

.github/workflows/build.yml
```diff
@@ -433,10 +401,15 @@ jobs:
           path: sboms/
 
       - name: Process CDX files
+        # Prevent direct interpolation of potentially untrusted Git ref values into the shell script:
+        # https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks
+        env:
+          REF_NAME: ${{ github.ref_name }}
```
peinjoh: Can we also add this REF_NAME extraction into build_context? We are doing it already as part of target-ref concat, so it would be just another re-export of the intermediate value.

peinjoh (approved):
Otherwise: LGTM

.github/workflows/build.yml
bshek-xitaso: Could you please review and merge https://github.com/csaf-rs/shared-workflows/pull/10 before merging this PR?

peinjoh: Done!

peinjoh (approved):

PR #802, opened 2026-08-12
feat: typed CVSS score-map constructors

gronke:
The typed CVSS accessors on `ContentTrait` (#715) cover the *read* direction for the untyped `cvss_v2`/`cvss_v3`/`cvss_v4` score maps. The *write* direction still forces consumers to hand-build `serde_json` maps. This adds the construction counterpart:

- `cvss_v2_to_score_map` / `cvss_v3_to_score_map` / `cvss_v4_to_score_map` render a typed `cvss-rs` model into the JSON object the property carries. (The `Cvss` enum itself has no `Serialize` impl; the per-version structs do, and their serde shape is exactly the FIRST schema's.)
- `cvss_v2_score_map_from_vector` / `cvss_v3_score_map_from_vector` / `cvss_v4_score_map_from_vector` parse a vector string and recompute the base score and severity band from the parsed metrics, so the mandatory consistency tests 6.1.9/6.1.10 hold *by construction* — an emitted document can never disagree with its own vectors.

The `cvss` module becomes public for this; the `validate_*` helpers it already carried become public API with it (happy to scope those `pub(crate)` instead if you'd rather keep the surface minimal).

## Motivation

A downstream SBOM/vulnerability tool emitting CSAF 2.0 currently rebuilds these JSON maps metric-by-metric to keep 6.1.9/6.1.10 green. With the typed models already in the tree, the crate can offer that construction once, for everyone.

## Notes

- Severity bands follow the FIRST specifications: v2 Low/Medium/High at 4.0/7.0; v3 and v4 None/Low/Medium/High/Critical at 0.0/4.0/7.0/9.0.
- A vector whose base metrics are incomplete keeps its parsed score as-is (`calculated_base_score` returns `None`).
- Error type is the shared `cvss_rs::ParseError` from the vector `FromStr` impls.

### DIFF
--- a/csaf-rs/src/cvss/mod.rs
+++ b/csaf-rs/src/cvss/mod.rs
@@ -343,3 +343,171 @@ pub fn check_optional_field_mismatch<T: PartialEq + std::fmt::Display + std::fmt
         _ => {},
     }
 }
+
+// ---- Typed construction of score objects ----
+
+use cvss_rs::v2_0::CvssV2;
+use cvss_rs::v3::CvssV3;
+use cvss_rs::v4_0::CvssV4;
+use serde::Serialize;
+use serde_json::Map;
+use std::str::FromStr;
+
+/// Renders a typed CVSS object into the JSON object that a score's
+/// `cvss_v2`/`cvss_v3`/`cvss_v4` property carries. The CSAF schemas type these
+/// properties as untyped objects; the typed models serialize to exactly the
+/// shape the referenced FIRST CVSS schemas expect.
+fn to_score_map<T: Serialize>(cvss: &T) -> Map<String, Value> {
+    match serde_json::to_value(cvss) {
+        Ok(Value::Object(map)) => map,
+        // A derived struct serialization always yields an object.
+        _ => Map::new(),
+    }
+}
+
+/// Renders a typed CVSS v2.0 object for a score's `cvss_v2` property.
+pub fn cvss_v2_to_score_map(cvss: &CvssV2) -> Map<String, Value> {
+    to_score_map(cvss)
+}
+
+/// Renders a typed CVSS v3.0/v3.1 object for a score's `cvss_v3` property.
+pub fn cvss_v3_to_score_map(cvss: &CvssV3) -> Map<String, Value> {
+    to_score_map(cvss)
+}
+
+/// Renders a typed CVSS v4.0 object for a content's `cvss_v4` property (CSAF 2.1).
+pub fn cvss_v4_to_score_map(cvss: &CvssV4) -> Map<String, Value> {
+    to_score_map(cvss)
+}
+
+/// Parses a CVSS v2.0 vector and renders its score object with the base score
+/// and severity recomputed from the parsed metrics, so the CSAF mandatory
+/// tests 6.1.9 and 6.1.10 hold by construction. A vector whose base metrics
+/// are incomplete keeps the parsed score as-is.
+pub fn cvss_v2_score_map_from_vector(vector: &str) -> Result<Map<String, Value>, cvss_rs::ParseError> {
+    let mut cvss = CvssV2::from_str(vector)?;
+    if let Some(score) = cvss.calculated_base_score() {
+        cvss.severity = Some(v2_severity(score));
+        cvss.base_score = score;
+    }
+    Ok(cvss_v2_to_score_map(&cvss))
+}
+
+/// Parses a CVSS v3.0/v3.1 vector and renders its score object with the base
+/// score and base severity recomputed from the parsed metrics, so the CSAF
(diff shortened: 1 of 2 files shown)


csaf-rs/src/cvss/mod.rs
```diff
@@ -343,3 +343,171 @@ pub fn check_optional_field_mismatch<T: PartialEq + std::fmt::Display + std::fmt
...
+    } else if score < 7.0 {
+        Severity::Medium
+    } else {
+        Severity::High
+    }
+}
+
+/// The CVSS v3.x severity band of a base score.
+fn v3_severity(score: f64) -> cvss_rs::v3::Severity {
+    use cvss_rs::v3::Severity;
```
aschierl-xitaso, line 436: It might make sense to use `map_score_to_severity` here, this is the existing implementation used for the tests and also cares about correct rounding.
(Looking at that, I just realized that test 6.1.9 does not yet check the severity for CVSS 2.0, #806, so the corresponding function for 2.0 does not yet exist)

csaf-rs/src/cvss/mod.rs
```diff
@@ -343,3 +343,171 @@ pub fn check_optional_field_mismatch<T: PartialEq + std::fmt::Display + std::fmt
...
+    if let Some(score) = cvss.calculated_base_score() {
+        cvss.base_severity = v4_severity(score);
+        cvss.base_score = score;
+    }
+    Ok(cvss_v4_to_score_map(&cvss))
+}
+
+/// The CVSS v2.0 severity band of a base score.
+fn v2_severity(score: f64) -> cvss_rs::v2_0::Severity {
+    use cvss_rs::v2_0::Severity;
```
aschierl-xitaso, line 424: We are currently in discussion, but the official first.org JSON schema for CVSS v2.0 does not include a "severity" field, so CSAF will likely reject documents that contain it. Severity will only be allowed starting with CVSS v3.

gronke:
Thanks for both comments, @aschierl-xitaso. Here is where I landed before pushing the next commits.

**CVSS v2.0 severity**

I checked the schema at https://www.first.org/cvss/cvss-v2.0.json to ground the decision. It defines 19 properties with `version`, `vectorString` and `baseScore` required; `severity` is not among them. As written the schema is draft-04 without `additionalProperties: false`, so the field is unspecified rather than strictly rejected, but that is thin ice to build on while the CSAF discussion is ongoing.

I will therefore drop the `severity` assignment from `cvss_v2_score_map_from_vector` and keep the recomputed `baseScore` only. `cvss_v2_to_score_map` on a caller-built model with `severity: None` already omits the key, so the crate stays within the FIRST v2 shape by default. If the discussion settles in favor of allowing severity, adding it back is a one-line change.

**Severity banding for v3/v4**

Agreed on reusing `map_score_to_severity`; it also gives the constructors the same scaled-integer rounding the tests use. The constructor and the validator must agree on bands, and keeping two implementations invites exactly the drift this PR tries to eliminate. It returns the unified `cvss_rs::Severity`, so v3/v4 get a small unified-to-version conversion (the inverse of `to_unified` in `cvss/v3.rs`). For v2 no mapping is needed once severity emission is dropped; when the v2 severity check for test 6.1.9 (#806) lands, its banding can live next to `map_score_to_severity` and this constructor can adopt it then.

**Missing required `version`**

While grounding the severity question in the schemas I noticed a gap in the current commits: all four FIRST schemas list `version` as required, but `CvssV2` and `CvssV4` in cvss-rs 0.4.0 carry no such field - it lives on the internally tagged `Cvss` enum, which is `Deserialize`-only. The v2/v4 score maps therefore serialize without `version` and would fail test 6.1.8 and `Cvss::deserialize` despite carrying correct scores. The next commits inject the constant into the rendered v2/v4 maps ("2.0"/"4.0"); v3 is unaffected on the vector path because `CvssV3::from_str` records the version from the mandatory prefix. Longer term the field arguably belongs in cvss-rs itself.

**Public surface of `mod cvss`**

One question back to you: the PR currently flips `pub(crate) mod cvss` to `pub mod cvss`, which also exposes the existing validation helpers (`validate_*`, `check_*`, the `create_*` error constructors, `deserialize_cvss`, `map_score_to_severity`) as public API. Would you rather keep the module fully public, or should I re-scope the validation internals to `pub(crate)` and export only the six new constructors? I slightly prefer the smaller surface, but the helpers are genuinely useful for custom validators, so I am happy with either.

tziemek:
I wouldn't expose the CVSS functionality here. Long term, these things should be part of cvss-rs itself.

gronke:
> I wouldn't expose the CVSS functionality here. Long term, these things should be part of cvss-rs itself.

Opened two (of a series of) PRs, proposing this to cvss-rs:

- [ ] scm-rs/cvss-rs/pull/40
- [ ] scm-rs/cvss-rs/pull/41

PR #804, opened 2026-08-13
build: don't benchmark when only markdown, scripts or build files change

tziemek:


peinjoh:
Minor/Question: Do we have an option to maintain the exclude list only in one place? For now (and <10?) entries this is fine, though 👍

peinjoh (approved):

PR #805, opened 2026-08-14
Verify CSAF 2.1 implementation of 6.2.11 and add supplementary tests

### LINKED ISSUE #297: Implement 6.2.11 Missing Canonical URL for CSAF 2.1
Already works for 2.0.
ToDo:
- [ ] update README
- [ ] update validation.rs

bshek-xitaso:
Resolves #297.

The existing implementation of 6.2.11 is compatible with CSAF 2.1. There are no semantic differences between the CSAF 2.0 and 2.1 specifications for this rule.

This PR:
- adds supplementary CSAF 2.1 test cases for previously uncovered scenarios
- marks 6.2.11 as implemented for CSAF 2.0 and 2.1 in the README

The CSAF 2.1 validation registration was already present, so no change to `validation.rs` is required.

The supplementary test cases cover the following scenarios:

Failing:
- no references - [s01](https://github.com/csaf-rs/csaf/pull/805/changes#diff-bbf87871dcfd1e4fa589f2b2566ce8dc8e4baeb0f4530ed8b65248643eae6419)
- an external reference category instead of self - [s02](https://github.com/csaf-rs/csaf/pull/805/changes#diff-a2ff123a07113f6e3707c02ec8af47d92a6a68ff3d3648f6005e5153d527fcff)
- an HTTP URL instead of HTTPS - [s03](https://github.com/csaf-rs/csaf/pull/805/changes#diff-44958c5fe60b4f6cd8af369a9805f25b430759827c5204c56b7c3d6c3e8e7b0a)
- multiple references where no single reference satisfies all canonical URL conditions - [s04](https://github.com/csaf-rs/csaf/pull/805/changes#diff-199370ab8df2df3aa55862e6a08a04a4874aa85ae863a6436620a3dda1ae7128)
- a missing reference category, which defaults to external - [s05](https://github.com/csaf-rs/csaf/pull/805/changes#diff-058ad00f416c8aebacca441c1ce1baf5fbb626d9fcacaa4ad95f875332f441fe)
- URLs with query parameters or fragments after the canonical filename - [s06](https://github.com/csaf-rs/csaf/pull/805/changes#diff-28f99ae3546b076af30bc8269dd84c8f1cb4de2a6121e97ca6eb6b02f8f1f31d)

Valid:
- one matching reference among non-matching references - [s11](https://github.com/csaf-rs/csaf/pull/805/changes#diff-2a2859a5a2712a41e58288c5fff74c760ef6f67ceb1bd01d0bb8ea087eaddd7d)
- canonical filename with multiple invalid character sequences in the tracking ID - [s12](https://github.com/csaf-rs/csaf/pull/805/changes#diff-57b47f4e51fae42742437fc0b42103ffe6ff7d459bb4e5904af5e8f0369c8b74)
- multiple matching references - [s13](https://github.com/csaf-rs/csaf/pull/805/changes#diff-9e5ad5bbd41ce5db4707bd4c3a53d665c6571def411c772bf822d0f611e7e56b)


type-generator/assets/tests/csaf_2.1/testcases.json
```diff
@@ -650,6 +650,46 @@
         }
       ]
     },
+    {
+      "id": "6.2.11",
+      "group": "recommended",
+      "failures": [
+        {
+          "name": "recommended/csaf-rs_csaf-csaf_2_1-6-2-11-s01.json",
+          "valid": true
```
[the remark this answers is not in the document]
peinjoh: This is fine, this test only generates warnings by definition.

csaf-rs/src/validations/test_6_2_11.rs
```diff
@@ -36,9 +36,22 @@ mod tests {
         // CSAF 2.0 has 2 test cases
```
peinjoh: Minor: This can be removed

csaf-rs/src/validations/test_6_2_11.rs
```diff
@@ -36,9 +36,22 @@ mod tests {
...
         TESTS_2_0.test_6_2_11.expect(err.clone(), ok.clone());
 
-        // CSAF 2.1 has 6 test cases
-        TESTS_2_1
-            .test_6_2_11
-            .expect(err.clone(), err.clone(), err, ok.clone(), ok.clone(), ok);
+        // CSAF 2.1 has 6 OASIS and 8 supplementary test cases (14 in total).
```
peinjoh: Minor: This can be removed

type-generator/assets/tests/csaf_2.1/testcases.json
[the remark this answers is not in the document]
tziemek: Wrong finding: valid indicates if the test against the `basic` preset has failed or not, not if this test case is failing, that is what `failures` is for. Everything fine here.

csaf-rs/src/validations/test_6_2_11.rs
```diff
@@ -36,9 +36,22 @@ mod tests {
...
+        TESTS_2_1.test_6_2_11.expect(
```
peinjoh: Minor: Could you clean up the comments a little more by moving the "What does what" description into a comment block before the fixture starts? And could you add similar comments for the existing passing/failing test cases?

Otherwise, I think this is a great improvement to the test coverage of this test!

It partly mirrors our internal unit tests for the "get canonical file name" and "is canonical url" functions, but as these test files are meant to be upstreamed and ideally used by other implementers, this is well and intended :)

Please open an issue to remind us that these test cases should be provided to upstream sometime soon!

peinjoh (approved):
LGTM 🚀

peinjoh:
Another thought: Could you add one more supplementary test that pins that http fragments / parameters does fail the 6.2.11 conditions? We are doing this correctly (I think) by simply checking that the URL ends with the substring, but another implementation could parse the string as some more advanced url type and then check against internal value there, which might have already parsed out the fragment / params. 

Additionally, and this might be a seperate chore PR: We usually provide the same test cases both as CSAF 2.0 and 2.1 test files, which is important for diverging requirements / implementations between versions. That is not the case here, so the PR is fine as-is. We should discuss if we want to keep duplicating the test files in these cases.

peinjoh (approved):
Looking good 👍

peinjoh:
If there is nothing left to do here @bshek-xitaso, I'll merge 👍

peinjoh:
Thank you @bshek-xitaso for your contributions so far, and for this first test task done 👍

PR #863, opened 2026-08-21
ci: use shared cargo publish dry-run action

bshek-xitaso:
This PR addresses https://github.com/csaf-rs/shared-workflows/issues/17.

It replaces the local `Cargo publish dry-run` step with the shared cargo-publish-dry-run composite action from [shared-workflows](https://github.com/csaf-rs/shared-workflows).


bshek-xitaso:
`cargo publish --dry-run` works here without `--workspace` because the root toml is a virtual workspace and no default-members are defined. In that case, Cargo automatically uses all workspace members, so the same as using `--workspace`.
https://doc.rust-lang.org/cargo/commands/cargo-publish.html#package-selection

peinjoh:
> `cargo publish --dry-run` works here without `--workspace` because the root toml is a virtual workspace and no default-members are defined. In that case, Cargo automatically uses all workspace members, so the same as using `--workspace`. https://doc.rust-lang.org/cargo/commands/cargo-publish.html#package-selection

Addition to this: In the csaf repo, only csaf-rs and csaf-validator get published allthough the workspace is initially selected as described. All other packages have `publish = false` in their cargo.toml files

peinjoh (approved):

PR #893, opened 2026-08-28
feat: add test 6.2.16 for CSAF 2.1

### LINKED ISSUE #302: Implement 6.2.16 Missing Product Identification Helper for CSAF 2.1
Already works for 2.0.
ToDo:
- [ ] update README
- [ ] update validation.rs

bshek-xitaso:
This PR resolves https://github.com/csaf-rs/csaf/issues/302.

It adds new supplementary test cases and tests the implementation of 6.2.16 for CSAF 2.1.

- s01 - multiple products in nested branches with alternating presence of product identification helpers
- s02 - multiple full product names with alternating presence of product identification helpers
- s03 - multiple product paths with alternating presence of product identification helpers
- s04 - multiple missing product identification helpers across all three product tree locations

- s11 - no product tree
- s12 - product tree with no relevant full product name types


peinjoh:
The test coverage is looking good, thank you for that!

One remark: What was the motivation to have the the supplementary test cases S03 and S12 be invalid in the context of mandatory tests? Both throw 6.1.1 errors (S03: 8, S04: 2 errors). If there this is not relevant / related to the test, I would kindly ask you to fix these before we merge. 

Otherwise LGTM!

Please also open an issue to complete the test coverage for CSAF 2.0, and an issue to upstream these test cases.

bshek-xitaso:
I assumed that it was acceptable for supplementary test cases to fail unrelated mandatory checks as long as they were marked with `valid: false` in `testcases.json`.

That's why I didn't add the referenced `full_product_names` to `s03`, since they weren't relevant to the `product_paths` test case. I intended this case to cover only `$.product_tree.product_paths[*].full_product_name`, and adding extra `full_product_names` to satisfy mandatory checks would only add content, unrelated to the `s03`'s test scope. That was the only reason for leaving them out, and I can add them back.

For `s12`, however, I can't make it mandatory-valid without changing the purpose of the test case. `product_tree` must contain at least one item to match the schema, and `product_groups` must reference existing products to pass the mandatory checks. Adding those products would introduce relevant `full_product_name` types and change what this case is meant to test. The idea of this test was: `product_tree` is present, but there are no relevant `full_product_name` types --> PASS (This is also why I added `product_groups` — it serves as a stub to satisfy the schema, but without the related products the file becomes mandatory-invalid).
Would you prefer me to keep `s12` as it is, or remove it?

Is it a convention that all supplementary test cases should also pass the mandatory tests?

peinjoh:
Thank you for the clarifications!

It is not a convention, more a "we try to make the tests only invalid in the way relevant to the test, within reason". I would argue adding the 3(?) missing full product name objs for s03 to make that file valid falls into that category, specially as it highlights the difference between FPNs with PIHs vs FPNs without PIHs. So I would kindly ask you to add them back. 

For s12, thats a really smart way to work around the json schema! I think this can be left as-is, as it highlights a real edge case where we don't want to report!

Please also summarize this discussion in the test doc comments and/or link your PR comment for posterity. 

Otherwise, thanks again for your contribution(s)!

peinjoh (approved):
LGTM, thank you for the contribution!

If this is ready from your side @bshek-xitaso, I'll merge

bshek-xitaso:
Yes, it is ready to be merged. Thank you for your time

peinjoh (approved):

PR #700, opened 2026-06-23
docs: add contributing guidelines

tziemek:
resolves #692

PR #906, opened 2026-09-01
feat: add test 6.2.7 for CSAF 2.1

### LINKED ISSUE #293: Implement 6.2.7 Missing Date in Involvements for CSAF 2.1
Already works for 2.0.
ToDo:
- [ ] update README
- [ ] update validation.rs

bshek-xitaso:
This PR resolves https://github.com/csaf-rs/csaf/issues/293.

It adds new supplementary test cases and tests the implementation of 6.2.7 for CSAF 2.1.

Failing supplementary test case:
- s01 - alternating presence of dates across multiple vulnerabilities and involvements (two vulnerabilities with four involvements each, alternating between present and missing dates to verify that all missing dates are detected regardless of whether they appear before or after a valid involvement, and that validation continues across all vulnerabilities and involvements after a finding)

- s11 - no vulnerabilities
- s12 - vulnerability without involvements
- s13 - vulnerability with one involvement containing a date


csaf-rs/src/validations/test_6_2_07.rs
```diff
@@ -44,11 +44,30 @@ mod tests {
     #[test]
     fn test_test_6_2_07() {
         let case_01 = Err(vec![create_missing_date_in_involvements_error(0, 0)]);
+        let case_s01 = Err(vec![
```
Gronner, line 47: As @peinjoh highlighting in my PR in the future speaking names should be used for the test cases. See also the upcoming contribution guidelines: https://github.com/csaf-rs/csaf/pull/700/changes#diff-fc045f5811272d70206705e0edc6ae36ddc392f1b798b945118db1a649cf6ea0R55

For now this is fine, as your comments provide the necessary information.

csaf-rs/src/validations/test_6_2_07.rs
```diff
@@ -44,11 +44,30 @@ mod tests {
...
         TESTS_2_0.test_6_2_7.expect(ExpectedResults_2_0 {
             case_01: case_01.clone(),
         });
-        TESTS_2_1.test_6_2_7.expect(ExpectedResults_2_1 { case_01 });
+
+        // Failing test cases:
+        // 01  - missing date in one involvement of one vulnerability
+        // s01 - alternating presence of dates across multiple vulnerabilities and involvements
+
+        // Valid test cases:
```
Gronner, line 62: After putting some more thought into this:

While s01 also covers "correct" entries, it is still a failing test. I think it makes sense to explicitly include one valid supplementary test that is just vulnerability[0].involvements[0].date.

I also took a look at what happens if the date field is an empty string: this is detected by the `schema` validations which are part of the `basic` tests, so besides a fully valid test I think this is good coverage.

Gronner (changes requested):

csaf-rs/src/validations/test_6_2_07.rs
bshek-xitaso: So, if I understand it correctly, either speaking names or explanatory comments are allowed, but speaking names are generally preferred?

Gronner: In the future it should be speaking names, for now comments is fine.

Gronner (approved):
LGTM

type-generator/assets/tests/csaf_2.1/recommended/csaf-rs_csaf-csaf_2_1-6-2-07-s01.json
```diff
@@ -0,0 +1,78 @@
...
+          "status": "in_progress"
+        },
+        {
+          "date": "2024-01-21T10:00:00.000Z",
+          "party": "vendor",
+          "status": "in_progress"
+        },
+        {
+          "party": "discoverer",
+          "status": "in_progress"
```
peinjoh: Comment, non blocking: While this is valid syntactically, this involvements history has some content-wise issues and does IMO no represent the correct usage of the involvements fields.

From the CSAF 2.1 spec:

> The value in_progress indicates that some hotfixes, permanent fixes, mitigations, workarounds, or patches may have been made available by the party, but more information or fixes may be released in the future. The use of this status by a vendor indicates that future information from the vendor about the vulnerability is to be expected.

This involvements list means:
The discoverer provided some fixes/workarounds/..., then the vendor did, then the coordinator also did, then the vendor did again.

A more "usual" flow could be: discoverer attempted contact, vendor provides fix, coordinator acknowledged via open, vendor put completed (which could be followed by coordinator / discoverer also putting completed)

I would argue we should represent a more "usual" involvement list here.

peinjoh:
Also, the usual reminder: Please add an issue for CSAF 2.0 back porting and for upstreaming the test cases.

peinjoh (approved):
Lets discuss the involvment history, otherwise LGTM

peinjoh (approved):
Involvements look better now, LGTM

PR #913, opened 2026-09-03
feat: add tests 6.2.4 for CSAF 2.1

### LINKED ISSUE #290: Implement 6.2.4 Build Metadata in Revision History for CSAF 2.1
Already works for 2.0.
ToDo:
- [x] check if current implementation works for 2.1
- [x] check and extend supplementary test cases for 2.1
- [x] update README
- [ ] update validation.rs

bshek-xitaso:
This PR resolves https://github.com/csaf-rs/csaf/issues/290.

It adds a new supplementary test cases and tests the implementation of 6.2.4 for CSAF 2.1. Since the rule is almost identical in CSAF 2.0, the supplementary test coverage is also reused for CSAF 2.0

- s01 - alternating presence of build metadata across multiple revisions (four semantic-versioned revisions alternating between values without and with build metadata to verify that all occurrences of build metadata are detected regardless of their position in the revision history, and that validation continues after a finding)
- s02 - build metadata following pre-release information (two pre-release semantic versions, one without and one with build metadata, to verify that build metadata is detected when it follows a pre-release part) (breaks 6.1.19, therefore - mandatory invalid, but represents an interesting case)

- s11 - one semantic-versioned revision without build metadata (counterpart to OASIS case 01)
- s12 - two revisions using integer versioning (without build metadata, because schema doesn't allow that)

### DIFF
--- a/csaf-rs/src/validations/test_6_2_04.rs
+++ b/csaf-rs/src/validations/test_6_2_04.rs
@@ -52,12 +52,12 @@ mod tests {
 
     #[test]
     fn test_test_6_2_04() {
-        let case_01_build_metadata = Err(vec![create_build_metadata_in_rev_history_error(
+        let revision_history_with_build_metadata = Err(vec![create_build_metadata_in_rev_history_error(
             &SemVerVersion::from(Version::from_str("1.0.0+exp.sha.ac00785").unwrap()),
             &0,
         )]);
 
-        let case_s01_mixed_build_metadata_presence = Err(vec![
+        let revision_history_with_alternating_build_metadata = Err(vec![
             create_build_metadata_in_rev_history_error(
                 &SemVerVersion::from(Version::from_str("1.1.0+exp.sha.ac00785").unwrap()),
                 &1,
@@ -68,22 +68,23 @@ mod tests {
             ),
         ]);
 
-        let case_s02_build_metadata_after_pre_release = Err(vec![create_build_metadata_in_rev_history_error(
-            &SemVerVersion::from(Version::from_str("1.1.0-rc.1+exp.sha.ac00785").unwrap()),
-            &1,
-        )]);
+        let revision_history_with_build_metadata_after_pre_release =
+            Err(vec![create_build_metadata_in_rev_history_error(
+                &SemVerVersion::from(Version::from_str("1.1.0-rc.1+exp.sha.ac00785").unwrap()),
+                &1,
+            )]);
 
         TESTS_2_0.test_6_2_4.expect(ExpectedResults_2_0 {
-            case_01: case_01_build_metadata.clone(),
+            case_01: revision_history_with_build_metadata.clone(),
         });
 
-        // Case s11: no build metadata
-        // Case s12: integer version numbers (no build metadata allowed by schema)
+        // Case s11: revision_history without build metadata
+        // Case s12: revision_history with integer versions (no build metadata allowed by schema)
 
         TESTS_2_1.test_6_2_4.expect(ExpectedResults_2_1 {
-            case_01: case_01_build_metadata,
-            case_s01: case_s01_mixed_build_metadata_presence,
-            case_s02: case_s02_build_metadata_after_pre_release,
+            case_01: revision_history_with_build_metadata,
+            case_s01: revision_history_with_alternating_build_metadata,
+            case_s02: revision_history_with_build_metadata_after_pre_release,
             case_s11: Ok(()),
             case_s12: Ok(()),
         });
--- a/csaf-rs/src/validations/test_6_2_04.rs
+++ b/csaf-rs/src/validations/test_6_2_04.rs
@@ -52,15 +52,40 @@ mod tests {
 
     #[test]
     fn test_test_6_2_04() {
-        let case_01 = Err(vec![create_build_metadata_in_rev_history_error(
+        let case_01_build_metadata = Err(vec![create_build_metadata_in_rev_history_error(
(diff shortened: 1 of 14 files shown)


csaf-rs/src/validations/test_6_2_04.rs
```diff
@@ -52,15 +52,40 @@ mod tests {
 
     #[test]
     fn test_test_6_2_04() {
-        let case_01 = Err(vec![create_build_metadata_in_rev_history_error(
+        let case_01_build_metadata = Err(vec![create_build_metadata_in_rev_history_error(
```
Gronner: I'd drop the case_01 prefix, as this information will be given by the Expected Result's member and maybe it might be reused in another expected result in the future. The same would apply to L60 and L71

I'd also suggest something like `revision_history_contains_build_metadata` or shorter `contains_build_metadata`.

csaf-rs/src/validations/test_6_2_04.rs
```diff
@@ -52,15 +52,40 @@ mod tests {
...
+
+        // Case s11: no build metadata
+        // Case s12: integer version numbers (no build metadata allowed by schema)
+
+        TESTS_2_1.test_6_2_4.expect(ExpectedResults_2_1 {
+            case_01: case_01_build_metadata,
+            case_s01: case_s01_mixed_build_metadata_presence,
+            case_s02: case_s02_build_metadata_after_pre_release,
+            case_s11: Ok(()),
+            case_s12: Ok(()),
```
Gronner: As the wording is almost identical across 2.1 and 2.0, please add them to 2.0 as well to increase test coverage.

Gronner (changes requested):

csaf-rs/src/validations/test_6_2_04.rs
bshek-xitaso: Good point. I [aligned](https://github.com/csaf-rs/csaf/pull/913/commits/df0e1e57c73312f92b8d73c7dbc90b89a3a8d40a) it a bit more closely with the naming style from our [Coding Guidelines](https://github.com/csaf-rs/csaf/pull/700/changes#diff-fc045f5811272d70206705e0edc6ae36ddc392f1b798b945118db1a649cf6ea0R55)

csaf-rs/src/validations/test_6_2_04.rs
bshek-xitaso: Done 👍

Gronner (approved):
LGTM

type-generator/assets/tests/csaf_2.0/optional/csaf-rs_csaf-csaf_2_0-6-2-04-s02.json
```diff
@@ -0,0 +1,31 @@
...
+          "summary": "Initial version."
+        },
+        {
+          "date": "2021-07-21T10:00:00.000Z",
+          "number": "1.1.0-rc.1+exp.sha.ac00785",
+          "summary": "Second version."
+        }
+      ],
+      "status": "draft",
+      "version": "1.1.0-rc.1"
```
peinjoh, line 28: Side node, non blocking: For S01, you correctly put the version prop without build metadata, as 6.1.16 ignores the build metadata anyway.

6.1.16 also ignores pre-release infos if the document status is draft, so you could have also put `1.1.0` here without failing 6.1.16.

type-generator/assets/tests/csaf_2.0/optional/csaf-rs_csaf-csaf_2_0-6-2-04-s02.json
```diff
@@ -0,0 +1,31 @@
...
+    "publisher": {
+      "category": "other",
+      "name": "CSAF-RS Test Files",
+      "namespace": "https://github.com/csaf-rs/csaf/tree/main/type-generator/assets/tests"
+    },
+    "title": "Optional test: Build Metadata in Revision History (failing supplementary example 2 - build metadata following pre-release versions)",
+    "tracking": {
+      "current_release_date": "2021-07-21T10:00:00.000Z",
+      "id": "CSAF-RS_CSAF-CSAF_2_0-6-2-04-S02",
+      "initial_release_date": "2021-07-21T10:00:00.000Z",
```
peinjoh, line 14: General comment, asking for change:`initial_release_date` does not need to be the first revision history items date, for example if multiple iterations were done before release to the specified target group.

If there is no indication of such a revision history, in the upstream test files like 6-1-14-01, the initial_release_date is set to the first revision history item. I think we should follow that pattern here, and I would kindly ask you to change the initial_release_date for consistency in all new test files.

csaf-rs/src/validations/test_6_2_04.rs
peinjoh, line 34: Also, side node: We should simplify that signature to usize, its the same size anyway.

type-generator/assets/tests/csaf_2.0/optional/csaf-rs_csaf-csaf_2_0-6-2-04-s02.json
bshek-xitaso: Thanks, that makes sense

csaf-rs/src/validations/test_6_2_04.rs
bshek-xitaso: Should I do that in a separate PR or include it in the same PR as the json changes?

peinjoh: I'd be fine with doing that in the same PR.

PR #929, opened 2026-09-07
refactor: 6.2.17 uses `Cve` instead of redefining regex

peinjoh:
Minor refactor:
We already have this regex defined via the schema, this removes the duplicate definition and uses the already existing regress regex. This also ensures that the definition of the CVE regex will stay up to date if the schema should change.


csaf-rs/src/validations/test_6_2_17.rs
```diff
@@ -21,7 +19,7 @@ pub fn test_6_2_17_cve_in_field_ids(doc: &impl CsafTrait) -> Result<(), Vec<Test
     for (v_i, vuln) in doc.get_vulnerabilities().iter().enumerate() {
         if let Some(ids) = vuln.get_ids() {
             for (i_i, id) in ids.iter().enumerate() {
-                if CVE_REGEX.is_match(id.get_text()) {
+                if Cve::from_str(id.get_text()).is_ok() {
```
Gronner, line 22: I've taken a second look at the regex in the specification/the one used here.
A simple ` CVE-2021-44228` (note the space at the beginning) would make this test pass and in its current state the validator running with `--test full` also does not throw any errors for the CVE created:

```text
/document/references: Document is missing a canonical URL [Warning 6.2.11]
 /document/lang: The document language is not defined [Warning 6.2.12]
 /vulnerabilities/0/ids/1/text: Vulnerability ID text 'CVE-2021-44229' matches CVE format [Warning 6.2.17]
 /vulnerabilities/0/ids/0/system_name: The system_name 'CVE Project' is not registered in RVISC. [Info 6.3.20]
 /vulnerabilities/0/ids/1/system_name: The system_name 'CVE Project' is not registered in RVISC. [Info 6.3.20]
```

The same output can be achieved by modifying `csaf-rs_csaf-csaf_2_1-6-2-17-s01.json` to include a leading space in the `text` field. Should we address this?

> text (text) contains a single unique label or tracking ID for the vulnerability.

As this field is not intended to contain free text, but should contain a unique label or tracking ID, maybe this can be resolved with a `str::trim()`?

peinjoh: The prose states: 
> It is sufficient to check, whether the property text matches the regex ^CVE-[0-9]{4}-[0-9]{4,}$.

Which pins the start / end of the string. I agree though, this should be trimmed / removed whitespace and invisible chars. Could be raised with TC though when we upstream the suppl test cases.

And I would prefer to have this in a seperate, easily revertible PR if TC disagrees on this.

Gronner: Fully agree, otherwise this change LGTM :+1:

Gronner (approved):

csaf-rs/src/validations/test_6_2_17.rs
peinjoh: Alright, will resolve here and open PR with the trim changeset

PR #930, opened 2026-09-08
fix: improve tokenization perf

peinjoh:
This PR:
* improves the tokenization performance of `tokenize_words`. Scanning the char count from the start of the text introduced a quadratic complexity, which could have been used to exhaust resources in service deployments (with the constraint that long texts need to be provided in the first place)


Gronner (approved):
LGTM

PR #933, opened 2026-09-08
feat: add script for test file generation

### LINKED ISSUE #918: Extract test file generation script (from #700)

tziemek:
This adds a script to generate supplementary test files. The docs will be in #700 
This is just to have the script before the other one is merged.


scripts/json/create_new_testfile.sh
```diff
@@ -0,0 +1,268 @@
...
+ASSETS_DIR="$SCRIPT_DIR/../../type-generator/assets/tests"
+
+usage() {
+  cat <<'EOF'
+Usage: create-new-testfile.sh --test <id> --validity <failure|valid> [--csaf-version <2.0|2.1|both>]
+
+Required:
+  -t, --test <id>        Test number from the specification, e.g. "6.1.29"
+  --validity <type>       Whether the new file is a "failure" (failing test case) or
+                          "valid" case.
```
Gronner: Maybe we should mirror the language used in the test_cases.json schema provided here: https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/test/validator/testcases_json_schema.json#L25

> States whether the test file is valid according to the CSAF standard.

scripts/json/create_new_testfile.sh
```diff
@@ -0,0 +1,268 @@
...
+# Normalize case type to "failure" or "valid"
+case "$CASE_TYPE" in
+  failure)
+    CASE_TYPE="failure"
+    ;;
+  valid)
+    CASE_TYPE="valid"
+    ;;
+  *)
+    echo "Error: --case must be one of failure/fail/invalid or valid/pass (got '$CASE_TYPE')" >&2
```
Gronner: How are `fail`/`invalid` and `pass` covered here? Wouldn't they be caught by the catch all?

scripts/json/create_new_testfile.sh
```diff
@@ -0,0 +1,268 @@
...
+    exit 1
+    ;;
+esac
+
+if ! [[ "$TEST_ID" =~ ^6(\.[0-9]+)+$ ]]; then
```
Gronner:
```suggestion
if ! [[ "$TEST_ID" =~ ^6(\.[0-9]+){2,3}$ ]]; then
```
Optional: This would also limit it closer to the actual set of possible test ids.

scripts/json/create_new_testfile.sh
```diff
@@ -0,0 +1,268 @@
...
+  local valid_bool="false"
+  if [[ "$CASE_TYPE" == "valid" ]]; then
+    case_description="valid"
+    valid_bool="true"
+  fi
+
+  # Create the test fixture from the template, updating its tracking id and
+  # title with a placeholder that should be edited afterwards.
+  jq --arg id "$tracking_id" \
+     --arg title "TODO: describe $case_description supplementary example for test $TEST_ID" \
```
Gronner, line 230: 👍

Gronner (changes requested):
Overall looks good to me, just a question regarding one case handling and a suggestion to narrow the regex.

scripts/json/create_new_testfile.sh
tziemek: They would be caught by the catch-all yes, I removed those and only have failure/valid now.

scripts/json/create_new_testfile.sh
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env bash
+#
+# create_new_testfile.sh
+#
+# Bootstraps a new supplementary CSAF test fixture under
+# type-generator/assets/tests/csaf_<version>/, wires it up in the
+# corresponding testcases.json, and regenerates the Rust test definitions.
+#
+# Usage:
+#   ./create_new_testfile.sh --test <id> --case <failure|valid> [--csaf-version <2.0|2.1|both>]
```
Gronner:
```suggestion
#   ./create_new_testfile.sh --test <id> --validity <failure|valid> [--csaf-version <2.0|2.1|both>]
```

Gronner (approved):
LGTM, minor docu change still left open

Gronner (approved):

PR #934, opened 2026-09-08
feat: add test 6.2.5 for CSAF 2.1

### LINKED ISSUE #291: Implement 6.2.5 Older Initial Release Date than Revision History for CSAF 2.1
Already works for 2.0.
ToDo:
- [ ] update README
- [ ] update validation.rs

bshek-xitaso:
This PR resolves https://github.com/csaf-rs/csaf/issues/291.

It adds supplementary test cases and tests the implementation of 6.2.5 for CSAF 2.1.

- s01 - the initial release date appears to be on the next day and year, but is actually older than the oldest revision (due to different timezones); similar to OASIS case 02, but extends the timezone coverage with an edge case where normalization crosses into the previous day and year.

- s11 - the initial release date is newer than the oldest revision; counterpart to OASIS case 01 (initial release date older than the oldest revision (basic negative)) and OASIS case 11 (initial release date equal to the oldest revision (basic equality positive)), covering the remaining basic positive case.
- s12 - initial release date and oldest revision use different positive and negative timezone offsets but represent the same instant after timezone normalization; this extends OASIS case 12 by checking equality when both the timestamps use different timezone offsets.
- s13 - the oldest revision is not the first entry in the revision history array and can only be selected correctly using timezone-aware ordering, verifying that the selection is independent of the "actual" array order; a case not covered by the existing OASIS test cases.


peinjoh:
LGTM, the timezone test cases are a bit repetitive, but fine with me. Props for the S13 test case, that might be really cool to provide to upstream! Thank you for this contribution!

peinjoh:
Also the usual: Please create issues to complete the coverage for CSAF 2.0 and to upstream the test cases.

peinjoh (approved):

tziemek (approved):

PR #954, opened 2026-09-09
fix: 6-1-30-S11 typo

peinjoh:
@tziemek The version being "3" violates 6.1.16 and should not matter to 6.1.30, correct?


Gronner (approved):
LGTM

tziemek:
> @tziemek The version being "3" violates 6.1.16 and should not matter to 6.1.30, correct?

3 would violate 6.1.16 yes and doesn't matter for 6.1.30. If you change to 2, make sure to check if the `valid` state in *testcases.json* is correct.

peinjoh:
3 would violate 6.1.16 yes and doesn't matter for 6.1.30. If you change to 2, make sure to check if the valid state in testcases.json is correct.

It is set to valid, which is why I found this in the first place. Alright, merging 🚀

PR #955, opened 2026-09-09
fix: 6-1-27-19-S11 uses csaf_base category

peinjoh:
csaf_security_advisory requires product tree and vulnerabilites to be present, S11 violated 6.1.27.4 and 11.

Any category except for csaf_superseded works for the test, and csaf_base does not have the above requirement


Gronner (approved):
LGTM

PR #965, opened 2026-09-10
feat: add test for 6.2.12 for CSAF 2.0 and 2.1

### LINKED ISSUE #298: Implement 6.2.12 Missing Document Language for CSAF 2.1
Already works for 2.0.
ToDo:
- [x] check if current implementation works for 2.1
- [x] check and extend supplementary test cases for 2.1
- [ ] update README
- [x] update validation.rs

Gronner:
Resolves #298


csaf-rs/src/validations/utils/raw_json.rs
```diff
@@ -0,0 +1,26 @@
+use serde_json::Value;
+
+/// Result enum specifing if and how a JSON property is present and set.
```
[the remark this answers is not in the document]
Gronner: Fixed 3 typos

csaf-rs/src/validations/utils/raw_json.rs
```diff
@@ -0,0 +1,26 @@
...
+    Empty,
+    /// Property is contained and has a non-empty value
+    Set,
+}
+
+/// Checks whether the property specified by `path` is present and set in the provided JSON data.
+/// Returns a [JsonValuePresence].
+pub(crate) fn is_present_and_set(path: &str, json: &Value) -> JsonValuePresence {
+    match json.pointer(path) {
+        Some(Value::Null) => JsonValuePresence::Unset,
```
peinjoh, line 19: TODO for me: Re-check if this Some(_) does cover only valid cases.

peinjoh: We should discuss if:
* Some(Value::Bool(t))
* Some(Value::Number(t))
Should count as set in the context of this test.

peinjoh (approved):
Lets discuss the bool / number edge case above, otherwise LGTM.

Since this is a first "way to handle preset and set" implementation, I would like a review from @tziemek and/or @aschierl-xitaso here too to get their opinions.

csaf-rs/src/validations/utils/raw_json.rs
Gronner: I've also thought about this, as the `json` crate handles `false` and `0` as "empty" (https://docs.rs/json/latest/json/enum.JsonValue.html#method.is_empty). While I can see some reasoning in this, I think looking at the JSON site of our things, using `0` and `false` as empty does not make that much sense, as there are sure valid uses cases for both.

Gronner: We will add a supplementary test case with "lang": 2, "lang": true. These should pass this test, but fail due to the schema for the validator.

tziemek (approved):

type-generator/assets/tests/csaf_2.0/testcases.json
```diff
@@ -645,6 +645,24 @@
...
           "name": "optional/csaf-rs_csaf-csaf_2_0-6-2-12-s11.json",
           "valid": true
+        },
+        {
+          "name": "optional/csaf-rs_csaf-csaf_2_0-6-2-12-s12.json",
+          "valid": true
+        },
+        {
+          "name": "optional/csaf-rs_csaf-csaf_2_0-6-2-12-s13.json",
+          "valid": true
```
peinjoh: Bool and number will validate the schema, should be valid = false I think, same for CSAF 2.1

PR #967, opened 2026-09-11
chore: align suppl test case publisher fields

peinjoh:
This PR:
* aligns the publisher fields of all supplementary test cases to match our expected values


Gronner (approved):
LGTM, but requires a JSON --fix

Issue #915, opened 2026-09-04
ci: add suppl test case id / namespace check


peinjoh:
All supplemental test cases should use 
```json
   "publisher": {
      "category": "other",
      "name": "CSAF-RS Test Files",
      "namespace": "https://github.com/csaf-rs/csaf/tree/main/type-generator/assets/tests"
    },
```

and a `tracking.id` in the format "CSAF-RS_CSAF-CSAF_[2_0/2_1]-6-[1/2/3]-[XX]-S[XX]" and a matching filename.

We should ensure on CI.

PR #971, opened 2026-09-11
fix: align supplemental test file tracking ids / filenames

peinjoh:
Part of #915 

This PR fixes malformatted tracking IDs in our testcases json files


Gronner (approved):
LGTM

PR #982, opened 2026-09-11
feat: add test 6.3.4 for CSAF 2.1

### LINKED ISSUE #272: Implement 6.3.4 Missing CWE for CSAF 2.1
Already works for 2.0.
ToDo:
- [ ] update README
- [ ] update validation.rs

peinjoh:
Resolves https://github.com/csaf-rs/csaf/issues/272

This PR:
* minor refactoring for early return on no vulnerabilities present
* test case for that
* wire up 6.3.4 for CSAF 2.1
* update README

### DIFF
--- a/csaf-rs/src/validations/test_6_3_04.rs
+++ b/csaf-rs/src/validations/test_6_3_04.rs
@@ -15,7 +15,14 @@ fn create_missing_cwe_error(vulnerability_index: usize, field_name: &str) -> Tes
 pub fn test_6_3_4_missing_cwe(doc: &impl CsafTrait) -> Result<(), Vec<TestFinding>> {
     let mut errors: Option<Vec<TestFinding>> = None;
 
-    for (v_i, vuln) in doc.get_vulnerabilities().iter().enumerate() {
+    let vulnerabilities = doc.get_vulnerabilities();
+
+    if vulnerabilities.is_empty() {
+        // TODO #409 wasSkipped
+        return Ok(());
+    }
+
+    for (v_i, vuln) in vulnerabilities.iter().enumerate() {
         if vuln.get_cwes().is_none() {
             errors
                 .get_or_insert_default()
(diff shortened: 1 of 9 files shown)


csaf-rs/src/validations/test_6_3_04.rs
```diff
@@ -15,7 +15,14 @@ fn create_missing_cwe_error(vulnerability_index: usize, field_name: &str) -> Tes
...
-    for (v_i, vuln) in doc.get_vulnerabilities().iter().enumerate() {
+    let vulnerabilities = doc.get_vulnerabilities();
+
+    if vulnerabilities.is_empty() {
+        // TODO #409 wasSkipped
+        return Ok(());
+    }
+
+    for (v_i, vuln) in vulnerabilities.iter().enumerate() {
         if vuln.get_cwes().is_none() {
```
Gronner, line 26: I don't think this is sufficient for passing the test (or at least its spirit).

Using 

```json
"vulnerabilities": [
  {
    "cve": "CVE-2019-0708",
    "cwes": [],
    "title": "BlueKeep"
  }
]
```
the test succeeds, I'd consider this a supplementary test that is important here.

On the other hand the schema validator remarks this:

```text
Validating file: empty_cwe_array
 /vulnerabilities/0/cwes: [] has less than 1 item [Error schema]
```

Gronner (changes requested):

csaf-rs/src/validations/test_6_3_04.rs
peinjoh: The test does not succeed if `cwes` is `[]`, as it has minItems: 1. Instead, the test never gets executed, as the validator does not deserialize schema-invalid json. I also considered this, as this will be relevant when we add lenient parsing.

Gronner (approved):
LGTM

PR #1008, opened 2026-09-15
fix: update canonical URL validation for 6.2.11

### LINKED ISSUE #297: Implement 6.2.11 Missing Canonical URL for CSAF 2.1

bshek-xitaso:
This PR resolves https://github.com/csaf-rs/csaf/issues/297.

It updates the canonical URL validation for rule 6.2.11 according to the changes introduced in [OASIS CSAF Editor Revision 2026-09-02](https://github.com/oasis-tcs/csaf/blob/editor-revision-2026-09-02/csaf_2.1/prose/share/csaf-v2.1-draft.md#6211-missing-canonical-url-) (non-empty hostname, filename preseded with "/").

The implementation has been checked against the newly added OASIS test cases 04 and 05, and both are rejected as expected. However, these test cases are currently only available in the latest version of the OASIS repository.

**Discussion Note:**
As discussed, the current hostname check should be considered an MVP. The implementation checks that the authority part between `https://` and the `path` is non-empty, but does not fully parse the URL authority.

Because of that, edge cases such as:

- `https://:443/file.json`
- `https://@:/file.json`
- `https://userinfo@/file.json`

could still pass this check even though they do not contain a valid non-empty hostname.

If stricter URL validation is needed, we should consider using an URL parser.


csaf-rs/src/csaf/traits/document_trait.rs
```diff
@@ -34,7 +34,22 @@ fn canonical_url_candidates<'a, R: ReferenceTrait>(
...
+
+            // Split the authority from the path
+            let Some((authority, path)) = after_scheme.split_once('/') else {
+                return false;
+            };
+
+            // Get the last path segment, which is the actual filename
+            let actual_filename = path.rsplit('/').next().unwrap_or(path);
+
+            !authority.is_empty() && actual_filename == expected_filename
```
peinjoh, line 51: minor: we can fail earlier here, i.e. we don't need to rsplit if authority is empty

csaf-rs/src/csaf/traits/document_trait.rs
```diff
@@ -34,7 +34,22 @@ fn canonical_url_candidates<'a, R: ReferenceTrait>(
...
+                return false;
+            };
+
+            // Split the authority from the path
+            let Some((authority, path)) = after_scheme.split_once('/') else {
+                return false;
+            };
+
+            // Get the last path segment, which is the actual filename
+            let actual_filename = path.rsplit('/').next().unwrap_or(path);
```
peinjoh, line 49: minor: this unwrap_or(path) hints that this could happen. rsplit is guaranteed to result at least one element though, so the first next() can be expect()'ed to always be Some()

peinjoh (commented):
During review, I've added some more test cases for the unit test coverage. I'll also provde a patch file to you for convenience.

```rust
   // empty
    #[case::no_references(None, 0)]
    // canonical urls
    #[case::match_simple_path(Some(vec![make_ref21("self", HTTPS_MATCH)]), 1)]
    #[case::match_well_known(Some(vec![make_ref21("self", HTTPS_MATCH_WELL_KNOWN)]), 1)]
    // non-canonical urls
    #[case::reference_category_external(Some(vec![make_ref21("external", HTTPS_MATCH)]), 0)]
    #[case::url_schema_http(Some(vec![make_ref21("self", HTTP_MATCH)]), 0)]
    #[case::url_schema_ftp(Some(vec![make_ref21("self",  "ftp://example.com/example-company-2019-yh3234.json")]), 0)]
    #[case::wrong_filename(Some(vec![make_ref21("self", "https://example.com/example-company-2019-yh3235.json")]), 0)]
    #[case::only_domain(Some(vec![make_ref21("self", "https://example.com")]), 0)]
    #[case::only_domain_with_delim(Some(vec![make_ref21("self", "https://example.com/")]), 0)]
    #[case::empty_authority(Some(vec![make_ref21("self", "https:///example-company-2019-yh3234.json")]), 0)]
    #[case::empty_authority_additional_path(Some(vec![make_ref21("self", "https:////example-company-2019-yh3234.json")]), 0)]
    #[case::filename_as_authority(Some(vec![make_ref21("self", "https://example-company-2019-yh3234.json")]), 0)]
    #[case::with_fragment(Some(vec![make_ref21("self", "https://example.com/example-company-2019-yh3234.json#fragment")]), 0)]
    #[case::with_param(Some(vec![make_ref21("self", "https://example.com/example-company-2019-yh3234.json?foo=1")]), 0)]
```

peinjoh:
> Because of that, edge cases such as:

    https://:443/file.json
    https://@:/file.json
    https://userinfo@/file.json

could still pass this check even though they do not contain a valid non-empty hostname.

Please also add those as a) unit test coverage with an accompanying note and b) as passing test cases.

csaf-rs/src/csaf/traits/document_trait.rs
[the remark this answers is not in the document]
bshek-xitaso: It was discussed, that the 2.0 version should also have the same requirements. I will create an issue for providing this idea to the upstream

bshek-xitaso: Yes, makes sense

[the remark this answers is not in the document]
bshek-xitaso: Done via https://github.com/csaf-rs/csaf/issues/1015

csaf-rs/src/csaf/traits/document_trait.rs
```diff
@@ -34,7 +35,31 @@ fn canonical_url_candidates<'a, R: ReferenceTrait>(
...
+
+            // Get the last path segment, which is the actual filename
+            let actual_filename = path
+                .rsplit('/')
+                .next()
+                .expect("rsplit() result is guaranteed to contain at least one element");
+
+            // Check if filenames match
+            actual_filename == expected_filename
+        })
```
Gronner: Looks good an readable to me 👍 Just to make you aware, for similar situation it might make sense to use `if let`-chains. but in this case readability suffers:

```rust
if let Some(after_scheme) = url.strip_prefix("https://") &&
       let Some((authority, path)) = after_scheme.split_once('/') {
        !authority.is_empty() && path.rsplit('/').next().expect(...) == expected_filename
} else {
    false
};
```

Alternatively `rsplit_once('/')` to make it explicit we don't care about the prefix

None of this is blocking, just a remark

Gronner (commented):
LGTM, just a remark. Otherwise I'd wait for peinjoh's approval after adding the test cases.

csaf-rs/src/csaf/traits/document_trait.rs
bshek-xitaso: Thank you, makes sense

peinjoh (approved):
LGTM, thank you for integrating the improved test coverage and adding the known limitations!

I'm with @Gronner here, rsplit_once is actually what we want, both in intention and performance. Please make that change, then we can merge 👍

peinjoh (changes requested):
On a second reading:

I think the known limitations are actually forbidden by the prose. The 4th constraint in the prose is: "The hostname is not empty.", in the comments, we describe the authority to be non-empty and check for that in the code.

Both in RFC 3986 and the HTTP RFC's, host is distinguished from username (http does not have that) / port, see the RFC 3986 ABNF: `authority = [ userinfo "@" ] host [ ":" port ]`. 

If we go of this, we should split / rsplit the authority again, and check if host is actually empty. The 3 limitations could then be used as more supplemental test cases.

Gronner:
> On a second reading:
> 
> I think the known limitations are actually forbidden by the prose. The 4th constraint in the prose is: "The hostname is not empty.", in the comments, we describe the authority to be non-empty and check for that in the code.
> 
> Both in RFC 3986 and the HTTP RFC's, host is distinguished from username (http does not have that) / port, see the RFC 3986 ABNF: `authority = [ userinfo "@" ] host [ ":" port ]`.
> 
> If we go of this, we should split / rsplit the authority again, and check if host is actually empty. The 3 limitations could then be used as more supplemental test cases.

At this point, would it make sense to include the `url` crate and let that handle the parsing?

peinjoh:
> At this point, would it make sense to include the `url` crate and let that handle the parsing?

We should probably benchmark the cost of that but yeah.

bshek-xitaso:
I decided to go with the manual non-empty hostname check: after stripping the optional userinfo and port parts from the authority, there must still be something left, which means the URL has a non-empty hostname. URI syntax itself is covered separately by the JSON Schema via `format: "uri"`.

tziemek (approved):

PR #1020, opened 2026-09-16
feat: implement rule 6.1.62 for CSAF 2.1

bshek-xitaso:


peinjoh:
Reviewing...

csaf-rs/src/validations/test_6_1_62.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    // Calculate the newest revision date if the revision history is present
+    // For document statuses other than "final" or "interim", the newest revision date is not relevant and will be None
+    let newest_revision_date = revision_history.as_mut().and_then(|revision_history| {
+        // Get sorted revision history and find the newest entry (panic if two entries are compared and either date is invalid)
+        revision_history.inplace_sort_by_date_then_number();
+
+        // Check if the revision history isn't empty
+        match revision_history.last() {
+            Some(revision) => match &revision.date {
```
peinjoh: We really ought to refactor the revision history. Out of scope, but this is really cumbersome.

tziemek: Extracting the latest revision date directly after `tracking.aggregate_revision_history()` would make it a little easier. But we could provide a helper function on the revision history itself, as we need this in more than one test case.

csaf-rs/src/validations/test_6_1_62.rs
```diff
@@ -0,0 +1,156 @@
...
+
+fn create_discovery_date_too_late_for_revision_error(
+    doc_status: &DocumentStatus,
+    discovery_date: &ValidCsafDateTime,
+    newest_revision_date: &ValidCsafDateTime,
+    vulnerability_index: usize,
+) -> TestFinding {
+    TestFinding::Error(TestFindingData {
+        message: format!(
+            "Discovery date ({discovery_date}) for vulnerability at index {vulnerability_index} is newer than the newest revision date ({newest_revision_date}) on a document with status {doc_status}"
```
tziemek: I think we use either ' or \`  for quoting within the message.
We can leave `for vulnerability at index.... ` out, this is clear through the `instance_path`.

Issue #1006, opened 2026-09-14
chore:  SSVC JSON Schema is now provided upstream


peinjoh:
We should pull it from the autoratiative source

PR #1028, opened 2026-09-18
refactor: use macro-generated JSON schema validator

### LINKED ISSUE #1027: Use compile-time generated validators from `jsonschema`
`jsonschema` recently introduced macros to generate validators at compile time.
First tests indicate that these lead to lower binary size, faster startup time and faster validation.

aschierl-xitaso:
resolves #1027


csaf-rs/assets/README.md
```diff
@@ -188,4 +191,18 @@ Terms of use according to https://cwe.mitre.org/about/termsofuse.html:
...
 > but not limited to loss of income, program interruption, loss of information, or > other pecuniary loss) arising 
 > out of use of this information, no matter the cause of action, even if The MITRE Corporation has been > advised 
-> of the possibility of such damages.
\ No newline at end of file
+> of the possibility of such damages.
+
+## Derived Assets
+
+Soem schema files are derived from the officials schemas, but with the modification that unevaluatedProperties are 
+not allowed, as needed for test 6.2.20.
```
Gronner, line 199:
```suggestion
Some schema files are derived from the official schemas, but with the modification that unevaluatedProperties are 
not allowed, as needed for test 6.2.20.
```

Gronner (commented):
I've not reviewed the provided JSON schemas, assuming they come from official sources. Rest looks good to me ✔️ .

Though a second member should also take a look

tziemek (approved):

peinjoh:
As part of this PR, the typegen arg was renamed. There are still 2 occurances of the old arg:

in asset-update.yml: 
`run: cargo run --locked -- --schema --test-schema --validation-schemas --rvisc --translations`
and in the type-generator README.md

This will break the asset-update.yml on the next run I believe

peinjoh (changes requested):
See comment above

peinjoh:
Similarly, the "are there relevant changes from typegen" check in build.yml checks only in /src. Now that changes are generated into /assets, not running the typegen after an asset schema update would not be found during build.

peinjoh:
Also, the now-upstream available selection list json schema (https://github.com/csaf-rs/csaf/issues/1006) is being loaded and strict-ified now. There is no entry in the assets README.md for it yet though.

I believe you can resolve #1006 after this :)

peinjoh (approved):

peinjoh:
Thanks for the fixes, LGTM 🚀

.github/workflows/asset-sync.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: Regenerate types from schema
         working-directory: type-generator
-        run: cargo run --locked -- --schema --test-schema --validation-schemas --rvisc --translations
+        run: cargo run --locked -- --schema --test-schema --strict-schemas --rvisc --translations
```
tziemek: Why do we explicitly pass these flags anyway? Or why don't we pass `--language-tags` and `--test-definitions`? 
If we omit all, then everything is built by default.

peinjoh: Agreed, we can probably omit all.

peinjoh (approved):

